PR #51, opened 2019-09-11
style bugs graph update

jomongeorge:


client/src/main.js
```diff
@@ -22,7 +22,7 @@ import App from './App.vue'
 import router from './router'
 import store from './store'
 import './filters'
-
+import * as nem2Sdk from 'nem2-sdk'
```
AnthonyLaw: I dont think we need this for here, we can import directly to file when we need it.

client/src/main.js
```diff
@@ -37,15 +37,16 @@ Vue.component('page-footer', Pagefooter)
 Vue.component('time-since', TimeSince)
 Vue.component('loader', w1)
 Vue.use(VueClipboard)
+Vue.use(nem2Sdk)
```
AnthonyLaw: we can import directly to file when we need it.

jomongeorge: yea I put this here for an experiment , removed it

client/src/main.js
jomongeorge: removed import * as nem2Sdk from 'nem2-sdk'

PR #87, opened 2019-09-19
Refactored the store

Alexhuszagh:
New features:
1. Store is now modular
2. Removed passed components to RecentBlocks, this is now a computed value.
3. Removed any `store` dispatch from the SDK methods, except as required in the listener.
4. Separated page and latest data functionality.
  - The `pageList` holds the 25 items in the current page being viewed.
  - The `latestList` holds the 25 most recent items declared on the block.
  - When the `pageIndex` is 0, the two track each other, otherwise, they do not track.
5. Actions are now actually actions.
  - Actions include things like `block/subscribe`, `block/initializePage`, `block/fetchNextPage` and `block/fetchPreviousPage`.
  - Previously, actions were glorified mutations, like `SET_BLOCKS_LIST`.
  - This refactoring simplifies extensibility and ease-of-use.


client/src/format.js
```diff
@@ -86,21 +83,17 @@ const formatAccount = accountInfo => {
 
 // FORMAT MOSAICS
 const formatMosaics = mosaics => {
-  mosaics.map(mosaic => {
+  return mosaics.map(mosaic => {
     mosaic.hex = mosaic.id.toHex()
     mosaic.amount = mosaic.amount.compact()
```
AnthonyLaw: I think it's missing `return mosaic` inside the `map()`

Alexhuszagh: Good point, thanks!

PR #90, opened 2019-09-20
fix getBlockInfoByHeight break

AnthonyLaw:


Alexhuszagh:
My fault, thanks for catching this.

PR #95, opened 2019-09-21
Refactored function

AnthonyLaw:
### Added
- getDiagnostic sdk function
- getmarketData api function
- chainInfo in chain store
- marketData in chain store
- marketData endpoint url

### DIFF
--- a/client/src/components/NemPriceGraph.vue
+++ b/client/src/components/NemPriceGraph.vue
@@ -19,8 +19,7 @@
 <template>
   <div class="widget m-0 z-1 nempricegraph_con bordr_rds_top0 pt-0 pb-0">
     <loader v-if="!loading"></loader>
-    <!-- <canvas id="nempricegraph" style="width:100%;height:180px"></canvas> -->
-    <div id="nempricegraph" style></div>
+    <ApexCharts type="area" :data="chartData" />
   </div>
 </template>
 <style>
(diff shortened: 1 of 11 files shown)


Alexhuszagh:
@AnthonyLaw Can you rebase it off the current master? This way, your commits don't include duplicated work.

client/src/components/NemPriceGraph.vue
```diff
@@ -66,137 +65,57 @@
...
-    width: 100%;
+  margin: 0px 0px 0px -10px;
+  width: 100%;
 }
 </style>
 <script>
-import axios from 'axios'
+import { mapGetters } from 'vuex'
+import w1 from '@/components/Chart.vue'
 import ApexCharts from 'apexcharts'
```
OlegMakarenko: We don't need this import anymore

client/src/components/NemPriceGraph.vue
```diff
@@ -66,137 +65,57 @@
...
-          axisBorder: {
-            show: false,
-            color: '#0998a6'
-          }
-        },
-        tooltip: {
-          enabled: true
-        }
+      if (this.marketData.historicalHourlyGraph.Data) {
+        this.marketData.historicalHourlyGraph.Data.forEach((item, index) => {
```
OlegMakarenko: I think it is better to move this part to the Store

client/src/components/NemPriceGraph.vue
OlegMakarenko: We don't need this import anymore

client/src/components/NemPriceGraph.vue
AnthonyLaw: will move to store :)

client/src/components/NemPriceGraph.vue
AnthonyLaw: thanks i will remove it!

PR #104, opened 2019-09-26
moved account info function to store

AnthonyLaw:
### Update

- moved account detail page function to `store/account`

### DIFF
--- a/client/src/store/account.js
+++ b/client/src/store/account.js
@@ -1,0 +1,121 @@
+/*
+ *
+ * Copyright (c) 2019-present for NEM
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License ");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


client/src/store/account.js
```diff
@@ -0,0 +1,121 @@
...
+  },
+  actions: {
+    // Fetch data from the SDK By Address.
+    async fetchAccountDataByAddress({ commit }, address) {
+      let accountInfo = await sdkAccount.getAccountInfoByAddress(address)
+      commit('setAccountInfo', accountInfo)
+
+      let accountBalance = []
+      accountInfo.mosaics.forEach((el, idx) => {
+        let temp = []
```
OlegMakarenko: Maybe it will be better to store this data in object instead of array. So we will have Array of Objects and it could be easily set into the TableListView

PR #122, opened 2019-10-11
Card enhancement, errors handle

OlegMakarenko:
- Loading animation added to **Card**

```
<Card :loading="loading"> 
</Card>
```
- Hide Card body slot while loading or error occurs 

- Dropdown filter added to **AccountDetail** transaction list
- Handling fetch error, infinity loading fix and error message added for:
1.  **TransactionDetail**
2.  **AccountDetail**

### DIFF
--- a/src/store/transaction.js
+++ b/src/store/transaction.js
@@ -110,7 +110,8 @@ export default {
     aggregateInnerTransactions: [],
     // Cosignatures Info for Transaction Type: Aggregate.
     aggregateCosignatures: [],
-    transactionInfoLoading: false
+    transactionInfoLoading: false,
+    transactionInfoError: false
   },
   getters: {
     getLatestList: util.getLatestList,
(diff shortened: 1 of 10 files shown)


src/store/transaction.js
```diff
@@ -146,6 +148,7 @@ export default {
     aggregateCosignatures: (state, aggregateCosignatures) => Vue.set(state, 'aggregateCosignatures', aggregateCosignatures),
     pageListFormatted: (state, pageListFormatted) => Vue.set(state, 'pageListFormatted', pageListFormatted),
     transactionInfoLoading: (state, v) => state.transactionInfoLoading = v,
+    transactionInfoError: (state, v) => state.transactionInfoError = v,
```
AnthonyLaw: should we use `Vue.set` to make everything synchronize?

OlegMakarenko: When we add a new object property, it is important to use Vue.set to make it reactive. `transactionInfoError` is already reactive and we change only it's value, so we can use `=` operator

PR #124, opened 2019-10-12
Update sdk 0.14 - compatible with Fushicho  

AnthonyLaw:
### Updated
- SDK updated to 0.14 
- Fushicho network compatible.
- network endpoint
- loaded mosaic detail and namespace detail from Endpoint

### Remove
- hardcored Mosaic and namespace DTP

### DIFF
--- a/src/config/index.js
+++ b/src/config/index.js
@@ -11,7 +11,7 @@ import {
 } from 'nem2-sdk'
 
 const PORT = '3000';
-let defaultNode = 'http://52.194.207.217:' + PORT;
+let defaultNode = 'http://172.105.115.19:' + PORT;
 
 
 if (localStorage['defaultNode'] && validURL(localStorage['defaultNode'])) {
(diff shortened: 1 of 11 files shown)


src/config/index.js
```diff
@@ -21,12 +22,13 @@ if (localStorage['defaultNode'] && validURL(localStorage['defaultNode'])) {
 export const Endpoint = {
   api: defaultNode,
   getCurrentNode: () => defaultNode.replace(/(^\w+:|^)\/\//, '').replace(':' + PORT, ''),
-  ws: 'ws://47.107.245.217:' + PORT,
+  ws: `ws://${DEFAULT_HOST}:${PORT}`,
   nodes: [
+    { protocol: 'http', domain: 'api-01.mt.us-west-2.nemtech.network', port: ':' + PORT },
```
OlegMakarenko: Maybe it is better to change `domain` to `host`, so it will has correlation with `DEFAULT_HOST`

src/infrastructure/getNamespace.js
```diff
@@ -53,34 +53,31 @@ class sdkNamespace {
...
+        let parent = namespaceNames.find(n => n.namespaceId.id.equals(namespace.parentId.id))
+        namespace.name = parent.name + '.' + namespace.name
+      }
+    })
 
-    // namespaceNames.map(namespace => {
-    //   namespace.namespaceId = namespace.namespaceId.toHex().toUpperCase()
-    //   namespace.name = namespace.name.toUpperCase()
-    // })
+    namespaceNames.map(namespace => {
```
OlegMakarenko: Why we call `namespaceNames.map()` twice?

src/views/AccountDetail.vue
```diff
@@ -3,9 +3,8 @@
         <div class="page-content-card-f">
 
 
-            <Card 
+            <Card
                 class="card-f card-full-width"
-                :loading="loading"
```
OlegMakarenko: I think we need loading animation here and you?

AnthonyLaw: i was missing loading in computed().

src/config/index.js
AnthonyLaw: good point.

PR #139, opened 2019-10-15
Fixed chunk asset wrong path #133

### LINKED ISSUE #129: Node switching not working

### LINKED ISSUE #133: Nginx configuration issue
**Expected Behaviour**

Browsing over to http://explorer.mt.nemtech.network/block/1 should work.

**Actual Behaviour**

Nginx configuration doesnt currently forward URI and produces 404.

Additionally, nginx doesnt redirect request to document_root as shown in following example:

Screenshot 2019-10-15 at 12 02 50 PM

**Environment**

http://explorer.mt.nemtech.network/block/1

OlegMakarenko:
- Fix #133 
> Additionally, nginx doesnt redirect request to document_root as shown in following example:
> 
> image
> 

- Fix #129 
- Config refactoring


src/config/dictionary/index.js
```diff
@@ -0,0 +1,3 @@
+export default {
+    
+}
```
AnthonyLaw: what is this dictionary ?

src/config/index.js
```diff
@@ -1,45 +1,13 @@
...
-    '((\\d{1,3}\\.){3}\\d{1,3}))' + // OR ip (v4) address
-    '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
-    '(\\?[;&a-z\\d%_.~+=-]*)?' + // query string
-    '(\\#[-a-z\\d_]*)?$', 'i') // fragment locator
-  return !!pattern.test(str)
-}
+export default {
+  pageMenu,
+  Endpoint
+}
```
AnthonyLaw: i think only 1 `export default` will do?

src/config/dictionary/index.js
OlegMakarenko: I will move [this part](https://github.com/nemfoundation/nem2-explorer/blob/efdc6369850bfb54483e163aec262d5202777106/src/store/ui.js#L25) here for different languages

src/config/index.js
OlegMakarenko: Destructuring import is not working with `export default` (but I guess I did it before) . So I needed to add named export here

OlegMakarenko: Destructuring import is not working with `export default` (but I guess I did it before) . So I needed to add named export here

src/config/dictionary/index.js
AnthonyLaw: may we can call it `i18n`

OlegMakarenko: Yes, that's better, thanks

PR #143, opened 2019-10-16
Adds support for the transactions, mosaics, and namespaces in the UI

Alexhuszagh:
Some of the extensive changes:
- Updates block transactions to use the cursor.
- Changed `Mosaic.vue` to `Mosaics.vue` and `Namespace.vue` to `Namespaces.vue`
- Changed `/mosaic` to `/mosaics` to be consistent with `/blocks` and `/transactions`.
- Changed `/namespace` to `/namespaces` to be consistent with `/blocks` and `/transactions`.
- Added TransactionTypeBox to store state.
- Added routes for all transactions based on cursors.
- Updated the Mosaics view to use real data based on cursors.
- Updated the Namespaces view to use real data based on cursors.


src/store/block.js
```diff
@@ -54,25 +56,30 @@ export default {
...
   mutations: {
-    setLatestList: util.setLatestList,
-    setPageList: util.setPageList,
-    setPageIndex: util.setPageIndex,
-    setSubscription: util.setSubscription,
-    setLoading: util.setLoading,
-    resetPageIndex: util.resetPageIndex,
-    addLatestItem(state, item) {
-      util.addLatestItemByKey(state, item, 'height', 1)
+    setLatestList: (state, list) => { state.latestList = list },
```
OlegMakarenko: May lose reactivity. For objects use `Vue.set(state, "latestList", list)` instead please

src/store/mosaic.js
```diff
@@ -16,25 +16,98 @@
...
+        // Page is loaded, need to fetch previous page.
+        const mosaic = pageList[0]
+        let mosaicList = await sdkMosaic.getMosaicsSinceIdWithLimit(util.PAGE_SIZE, mosaic.id)
+        commit('setPageIndex', pageIndex + 1)
+        commit('setPageList', mosaicList)
+      }
+      commit('setLoading', false)
+    },
+
+    // Reset the block page to the latest list (index 0)
```
AnthonyLaw: should be reset the `mosaics page` instead of `block page`

src/store/namespace.js
```diff
@@ -16,33 +16,102 @@
...
+        // Page is loaded, need to fetch previous page.
+        const namespace = pageList[0]
+        let namespaceList = await sdkNamespace.getNamespacesSinceIdWithLimit(util.PAGE_SIZE, namespace.id)
+        commit('setPageIndex', pageIndex + 1)
+        commit('setPageList', namespaceList)
+      }
+      commit('setLoading', false)
+    },
+
+    // Reset the block page to the latest list (index 0)
```
AnthonyLaw: should be reset the `namespaces page` instead of `block page`

src/store/ui.js
```diff
@@ -77,6 +78,7 @@ export default {
       'startHeight': 'Start Height',
       'endHeight': 'End Height',
       'divisibility': 'Divisibility',
+      'supply': 'Supply',
```
AnthonyLaw: duplicated `supply` key and value

src/components/TransactionRow.vue
```diff
@@ -25,7 +25,7 @@
     <td>
       <AddressLink
         v-if="item.transactionBody.recipient"
-        :address="item.transactionBody.recipient.address"
+        :address="item.transactionBody.recipient"
```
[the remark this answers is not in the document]
AnthonyLaw: `transactionBody` is existed in every transactions.

src/format.js
```diff
@@ -122,22 +122,23 @@ const formatMosaics = mosaics => {
...
-    revision: mosaicInfo.revision,
-    startHeight: mosaicInfo.height.compact(),
-    duration: mosaicInfo.duration.compact(),
-    supplyMutable: mosaicInfo.flags.supplyMutable,
-    transferable: mosaicInfo.flags.transferable,
-    restrictable: mosaicInfo.flags.restrictable
-  }
-
-  return mosaicObj
+const formatMosaicInfo = mosaicInfo => ({
```
AnthonyLaw: i think should include with mosaicName as well.
so we can get MosaicIdHex and which namespace there are belong.

src/infrastructure/getMosaic.js
```diff
@@ -46,19 +50,55 @@ class sdkMosaic {
     return mosaicAmount
   }
 
-  static getMosaicInfo = async mosaicHexOrNamespace => {
-    let mosaicID = ''
+// TODO(ahuszagh) Remove...
+//  static getMosaicInfo = async mosaicHexOrNamespace => {
```
AnthonyLaw: IMO, we may need to accept `mosaicHex` or `Namespace` as params.
example `308f144790cd7bc4` is mosaicHex.

but some mosaic are link with then `namespace` such as `nem.xem`

so the params should accept `nem.xem` and convert to `308f144790cd7bc4`, then only request `/mosaics/from/${mosaicId}/limit/${limit}`

src/infrastructure/getNamespace.js
```diff
@@ -76,6 +79,40 @@ class sdkNamespace {
 
     return format.formatNamespace(namespaceInfo, namespaceNames)
   }
+
+  static getNamespacesFromIdWithLimit = async (limit, fromNamespaceId) => {
```
AnthonyLaw: same thing as mosaic.

it should able accept `namespaceId` or `namespace name`

src/components/TransactionRow.vue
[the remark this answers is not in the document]
Alexhuszagh: @OlegMakarenko, it exists in every case, but the recipient does not. Also, the recipient is now a formatted string, and no longer an object with an `address` field.

src/components/TransactionTypeBox.vue
```diff
@@ -21,14 +21,46 @@
...
       <div class="select_type">
-        <select>
-          <option>Recent Transactions</option>
-          <option>Pending Transactions</option>
-          <option>Transfer Transactions</option>
-          <option>Multisig Transactions</option>
-          <option>Mosaic Transactions</option>
+        <select @change="onChange($event)">
+            <option
+                v-for="option in Object.keys(transactionTypeMap)"
```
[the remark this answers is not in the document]
Alexhuszagh: @OlegMakarenko Right, but I only want the object keys, not the values. I'm generally a practitioner of "don't name what you don't use".

src/store/mosaic.js
Alexhuszagh: Yes, thanks for catching that.

src/format.js
Alexhuszagh: @AnthonyLaw That's not in the mosaic info model, however. We should have a function named `formatMosaicNames` or something similar, since the mosaic names uses something very different than the `MosaicInfo` model. In the `Mosaics` view, this would require a separate API call currently to get he name for each mosaic, and it's also fairly convoluted code to read (would take me a few days to actually get it running).

src/infrastructure/getMosaic.js
Alexhuszagh: @AnthonyLaw We're going to be paginating this from a list of mosaic info models, not mosaic names. We have the ID readily available, getting the name requires a lot of spurious requests.

src/components/TransactionTypeBox.vue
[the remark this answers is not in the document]
Alexhuszagh: @OlegMakarenko, Thanks for catching this.

src/store/block.js
Alexhuszagh: @OlegMakarenko, aren't we only calling this with `commit`, which should ensure reactivity regardless? Vue seems to not require `Vue.set` for mutations in the store, since those are reactive by nature.

OlegMakarenko: @Alexhuszagh Sorry, my bad. Yes, in this case you are right. Refresh whole object with new one - it works. There is issue with mutating object properties in Vuex Store

evias:
also please rebase to master since yesterday's merge(s) :+1:

Issue #134, opened 2019-10-15
Empty lists should have empty message


evias:
**Expected Behaviour**

Empty lists should show a message :

- `No transactions to display for this account`
- `No transactions to display for this block`
- `No namespaces to display for this account`
- etc.

PR #147, opened 2019-10-18
TableView: message added for empty data

OlegMakarenko:
#134 

- When passing empty Object/Array to TableView component it will show the message: "Nothing to show.."

- Also this message can be customized:
```
<TableListView
       :data="blockTransactionList"
       :emptyDataMessage="'No transactions to display for this block'"
/>
```

### DIFF
--- a/src/views/AccountDetail.vue
+++ b/src/views/AccountDetail.vue
@@ -11,7 +11,7 @@
         <template #error>Account {{address}} is not exist</template>
       </Card>
 
-      <Card v-if="1||showMultiSigDeatail" class="card-f card-adaptive" :loading="loading">
+      <Card class="card-f card-adaptive" :loading="loading">
         <!-- MultiSig Info -->
         <template #title>Multsig Info</template>
 
@@ -20,7 +20,7 @@
         </template>
       </Card>
 
-      <Card v-if="1||showMultiSigDeatail" class="card-f card-adaptive" :loading="loading">
+      <Card class="card-f card-adaptive" :loading="loading">
         <!-- MultiSig Cosignatories -->
         <template #title>Multsig Cosignatories</template>
 
@@ -30,7 +30,7 @@
 
       </Card>
 
-      <Card v-if="1||showMosaics" class="card-f card-adaptive" :loading="loading">
+      <Card class="card-f card-adaptive" :loading="loading">
         <!-- Mosaics -->
         <template #title>Owned Mosaics</template>
 
@@ -39,7 +39,7 @@
         </template>
       </Card>
 
-      <Card v-if="1||showNamespaces" class="card-f card-adaptive" :loading="loading">
+      <Card class="card-f card-adaptive" :loading="loading">
         <!-- NS -->
         <template #title>Owned Namespaces</template>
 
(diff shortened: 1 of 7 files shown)


src/views/AccountDetail.vue
```diff
@@ -11,7 +11,7 @@
         <template #error>Account {{address}} is not exist</template>
       </Card>
 
-      <Card v-if="showMultiSigDeatail" class="card-f card-adaptive" :loading="loading">
+      <Card v-if="1||showMultiSigDeatail" class="card-f card-adaptive" :loading="loading">
```
AnthonyLaw: what is the `1` meaning?

OlegMakarenko: Thanks. Missed to delete it. I use this as a "comment" for "if".

src/views/AccountDetail.vue
```diff
@@ -11,7 +11,7 @@
         <template #error>Account {{address}} is not exist</template>
       </Card>
 
-      <Card v-if="1||showMultiSigDeatail" class="card-f card-adaptive" :loading="loading">
```
AnthonyLaw: i think `showMultiSigDeatail` is valild because it existing in `computed` , only `1||` have to remove.

src/views/AccountDetail.vue
```diff
@@ -20,7 +20,7 @@
         </template>
       </Card>
 
-      <Card v-if="1||showMultiSigDeatail" class="card-f card-adaptive" :loading="loading">
```
AnthonyLaw: i think `showMultiSigDeatail` is valild because it existing in `computed` , only `1||` have to remove.

src/views/AccountDetail.vue
```diff
@@ -30,7 +30,7 @@
 
       </Card>
 
-      <Card v-if="1||showMosaics" class="card-f card-adaptive" :loading="loading">
```
AnthonyLaw: i think `showMosaics ` is valild because it existing in `computed` , only `1||` have to remove.

src/views/AccountDetail.vue
```diff
@@ -39,7 +39,7 @@
         </template>
       </Card>
 
-      <Card v-if="1||showNamespaces" class="card-f card-adaptive" :loading="loading">
```
AnthonyLaw: i think `showNamespaces ` is valild because it existing in `computed` , only `1||` have to remove.

src/views/AccountDetail.vue
```diff
@@ -48,7 +48,7 @@
         </template>
       </Card>
 
-      <Card v-if="1||showTransactions" class="card-f card-full-width" :loading="loading">
```
AnthonyLaw: i think `showTransactions ` is valild because it existing in `computed` , only `1||` have to remove.

src/views/AccountDetail.vue
OlegMakarenko: Referring the issue #134 all this cards should be shown even there is no items in a list. It is the reason I removed `v-for`s. I havent removed computed props linked with that, so we can go back to this`v-for` approach if needed

AnthonyLaw: i see, got it

Issue #146, opened 2019-10-18
Add node list to configuration file


OlegMakarenko:
Allow to overwrite the list of default nodes if a config file is provided

PR #151, opened 2019-10-22
Endpoint class moved to store. Node list fetch from config

OlegMakarenko:
#146 

- Endpoint class moved to store
- Endpoints (`marketDataUrl`) moved to `src/config/endpoints.json`
- Added init functions to infrastructure
- Call initSdk from store 
- Errors handle on root init
- Default nnode list moved to `src/config/peers-api.json`
- Client calls get requset `/config/peers-api.json` and rewrites default nodes if json provided
- Refresh store after node changed (not a page refresh)
- Hide NodeSelector dropdown after node selected

`Endpoint.api` => `store.getters['api/currentNode'].url`


src/helper.js
```diff
@@ -80,14 +80,35 @@ class helper {
...
+          hostname: url.hostname,
+          port: url.port,
+          url: rawUrl
+      }
+    }
+  }
+
+  static httpToWsUrl(url) {
+    if(this.validURL(url)){
+      return url.replace('http', 'ws')
```
AnthonyLaw: `.replace(/(http)(s)?\:\/\//, "ws$2://");`

it should support https as well.

OlegMakarenko: It supports https. Just leaves 's' at the end and replace http with ws

AnthonyLaw: my bad, sorry.

Alexhuszagh:
@OlegMakarenko We currently require the Endpoint for a few different locations:

1. https://github.com/nemfoundation/nem2-explorer/blob/master/src/components/PageFooter.vue#L69
2. https://github.com/nemfoundation/nem2-explorer/blob/a18a3f41608991804b1d6e78e89ea148d81a106b/src/components/TopHead.vue#L80
3. https://github.com/nemfoundation/nem2-explorer/blob/f194777ea5bbe0255233a4e978ca6fe3ec963872/src/store/ui.js#L174

PR #155, opened 2019-10-23
Fix #153 by fully removing the Endpoint.

### LINKED ISSUE #153: Insufficient Endpoint Removal
Endpoint, which was removed in #151, is still referenced in the following locations:

1. https://github.com/nemfoundation/nem2-explorer/blob/master/src/components/PageFooter.vue#L69
2. https://github.com/nemfoundation/nem2-explorer/blob/a18a3f41608991804b1d6e78e89ea148d81a106b/src/components/TopHead.vue#L80
3. https://github.com/nemfoundation/nem2-explorer/blob/f194777ea5bbe0255233a4e978ca6fe3ec963872/src/store/ui.js#L174

We need to remove these references to the endpoint class.

Alexhuszagh:
Patches the insufficient removal of Endpoint.

### DIFF
--- a/src/store/ui.js
+++ b/src/store/ui.js
@@ -16,8 +16,8 @@
  *
  */
 import router from '../router'
-import { Address, AccountHttp } from 'nem2-sdk'
-import { Endpoint, i18n } from '../config'
+import { Address } from 'nem2-sdk'
+import { i18n } from '../config'
 
 export default {
   namespaced: true,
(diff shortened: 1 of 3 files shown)


src/store/ui.js
```diff
@@ -16,8 +16,8 @@
  *
  */
 import router from '../router'
-import { Address, AccountHttp } from 'nem2-sdk'
-import { Endpoint, i18n } from '../config'
+import { Address } from 'nem2-sdk'
```
OlegMakarenko: Missed `AccountHttp`

Alexhuszagh: @OlegMakarenko Thanks, good catch.

PR #159, opened 2019-10-23
Patch the HTTP methods to use the current node from the store.

Alexhuszagh:
1. Removes the hacky `initializeSdk` methods.
2. Creates a helper class `http` with getters that create a wrapper to the client using the current node from the store.
3. Implemented all routes in terms of this helper.

### DIFF
--- a/src/infrastructure/http.js
+++ b/src/infrastructure/http.js
@@ -1,0 +1,66 @@
+/*
+ *
+ * Copyright (c) 2019-present for NEM
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License ");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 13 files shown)


src/infrastructure/http.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import store from '../store'
```
OlegMakarenko: Does it works? It seems store will be undefined

Alexhuszagh: @OlegMakarenko It works. Otherwise, we need to pass the node URL to each method or to a class instance.

PR #165, opened 2019-10-24
No data in TableView after page refresh fix

OlegMakarenko:


src/main.js
```diff
@@ -17,23 +17,20 @@
...
-
 import store from './store'
-window.store = store
-
 import TopHead from '@/components/TopHead.vue'
-import PageMenu from '@/components/PageMenu.vue'
+import PageMenu from './components/menu/PageMenu.vue'
 import Pagefooter from '@/components/PageFooter.vue'
 import TimeSince from '@/components/TimeSince.vue'
 import w1 from '@/components/Loading.vue'
```
AnthonyLaw: can just rename it `w1` as well?

OlegMakarenko: Sure

PR #167, opened 2019-10-25
Add timelines

Alexhuszagh:
1. Simplifies the stores to have a reusable `Timeline` class.
2. Adds a Button `Less` when no more data can be retrieved.
3. Removed `pageIndex` and changed it to t`imeline.index`


src/components/controls/Pagination.vue
```diff
@@ -1,6 +1,23 @@
 <template>
   <div class="pagination-wrapper">
-    <ButtonMore v-if="pageIndex === 0" @click="nextPage">{{ buttonText }}</ButtonMore>
+    <div v-if="!canFetchPrevious && !canFetchNext">
```
OlegMakarenko: We show pagination when cannot fetch prev and next?

src/store/timeline.js
```diff
@@ -0,0 +1,86 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+export default class Timeline {
+  static get pageSize() {
+    return 25
```
OlegMakarenko: It also can be moved to config I think

Alexhuszagh: @OlegMakarenko, sounds good, I can remove a lot of dead code and move that to a config shortly.

PR #211, opened 2019-11-07
Add accounts view.

Alexhuszagh:
Adds the account view.

Still needs work:
1). Needs the balance (acquired from mosaics).
2). Needs to last activity (acquired from the activity buckets).


Alexhuszagh:
@AnthonyLaw Thanks, good catch.

PR #220, opened 2019-11-11
Fixed initialization to minimize number of requests.

Alexhuszagh:
## Summary

1. Memoized the network type to minimize number of requests.
2. Removed global `initialize`, and changed to request data only as-needed.
3. Each view calls `api/initialize`, the their own initializer.
4. Simplified the detail views to use global `View`.
5. Ensured detail views use `loading` and `error` with `Card`.
6. Added errors for the timeline requests.
7. Ensured timeline views using `error` with `Card`.
8. Unified formatting for views.
9. Fixes 'Last Activity' not being clickable for accounts.

## Detailed

This fixes a lot of errors, formatting issues, and other potential issues:

1. Avoids 503 (Service Unavailable) errors by only requesting the minimum amount of data and memoizes the network type to avoid re-requesting it from the same node redundantly.
  - Previously, requesting info from a page other than home would request all the home data, request the data for the page, and make redundant requests to `/node/info` for each request for the network type.
  - Now, the global `initialize` has been removed, and `api/initialize` is required for each request, and only the data for each page is required.
  - `http.networkType` has been memoized and used for each network type request, and changes when the node changes. This allows us to avoid redunant requests and incurring 429 (Too Many Requests) or 503 (Service Unavailable) errors from the REST API.
2. Removed global `initialize`.
  - Data for the home page was being requests in the global initializer, which was not needed for each individual page. Now, each page only requests data needed for their view.
3. Each view calls `api/initialize` and their own initializer.
  - Rather than request data for the homepage, each view requests the minimum data required, only necessitating the API has been initialized.
4. Simplified `detail` views to use global `View`.
  - `NamespaceDetail` and `MosaicDetail` were using old components, which have been re-factored to properly use the new `Card` and `View` components.
5. Ensured detail views use `loading` and `error` with `Card`.
  - Ensured all the detail views have `error` with the first `Card` usage on the view, and `loading` for all `Card` usages on the view.
6. Added errors for the timeline requests.
  - Added `error` to `state` for each store module, `getError`, and `setError` for the getters and mutations.
  - Wrapped each timeline API request with a `try/catch` to ensure any errors are properly handled.
7. Ensured timeline views using `error` with `Card`.
  - Ensured all the timeline views (`Accounts`, `Blocks`, etc.) use `Card` with `error`, and have an error message if the data is unable to be fetched.
8. Unified formatting for views.
  - Updated formatting to be consistent with @OlegMakarenko's style for the views for older components and views.
  - Provides visual consistency with the rest of the codebase.
9. Fixes 'Last Activity' not being clickable for accounts.
  - Made `lastActivity` clickable.
  - Linked `lastActivity` to the `BlockDetail`.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -40,7 +40,7 @@ export default {
   },
   methods: {
     initialize() {
-      this.$store.dispatch('initialize')
+      this.$store.dispatch('api/initialize')
         .catch(error => console.log(error))
     },
     uninitialize() {
--- a/src/components/NemPriceGraph.vue
+++ b/src/components/NemPriceGraph.vue
@@ -70,8 +70,9 @@
 }
 </style>
 <script>
-import { mapGetters } from 'vuex'
 import w1 from '@/components/Chart.vue'
+import helper from '../helper'
+import { mapGetters } from 'vuex'
 
 export default {
   components: {
@@ -95,7 +96,10 @@ export default {
       marketData: 'chain/getMarketData'
     })
   },
-  mounted() {},
+  async mounted() {
+    await helper.logError(this.$store.dispatch, 'api/initialize')
+    await helper.logError(this.$store.dispatch, 'chain/initialize')
+  },
   methods: {}
 }
 </script>
--- a/src/components/RecentBlocks.vue
+++ b/src/components/RecentBlocks.vue
@@ -17,39 +17,64 @@
  */
 
 <template>
-  <div class="widget has-shadow">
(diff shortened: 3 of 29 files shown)


Alexhuszagh:
> Awesome, i can feel the page is load faster than before +1
> maybe just my feeling

Definitely, since it means less requests are required, so the page loads smoother and with less overhead. Also, the loading of each view other than home is a fairly drastic difference.

PR #226, opened 2019-11-13
Updated Nodes to use pseudo-dynamic data.

Alexhuszagh:
1. Create a store for the nodes.
2. Added hard-coded data to a store for the nodes list.
3. Simulated fetching data with our models from an API.
4. Re-implemented the `Nodes` view using the store.
5. Renamed various router routes for consistency.
6. Added `node` to `store` and added initialization callback.
7. Added terms in `en-us.json` for missing key names.

### DIFF
--- a/src/store/node.js
+++ b/src/store/node.js
@@ -1,0 +1,188 @@
+/*
+ *
+ * Copyright (c) 2019-present for NEM
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License ");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


src/store/node.js
```diff
@@ -0,0 +1,188 @@
...
+    getError: state => state.error
+  },
+  mutations: {
+    setInitialized: (state, initialized) => { state.initialized = initialized },
+    setTimeline: (state, timeline) => { state.timeline = timeline },
+    setLoading: (state, loading) => { state.loading = loading },
+    setError: (state, error) => { state.error = error }
+  },
+  actions: {
+    // Initialize the mosaic model.
```
OlegMakarenko: `mosaic` -> `node`

src/store/node.js
```diff
@@ -0,0 +1,188 @@
...
+    async initialize({ commit, dispatch, getters }) {
+      const callback = async () => {
+        await dispatch('initializePage')
+      }
+      await LOCK.initialize(callback, commit, dispatch, getters)
+    },
+
+    // Uninitialize the mosaic model.
```
OlegMakarenko: `mosaic` -> `node`

Alexhuszagh: @OlegMakarenko Fixed, nice catch.

Issue #246, opened 2019-11-20
Transaction page uses different table component


OlegMakarenko:
It needs to be the same table component in each page

![image](https://user-images.githubusercontent.com/33131259/69238626-1c42c700-0ba1-11ea-978a-ab23d98bd3cd.png)

**Environment**
http://explorer.mt.nemtech.network/transactions

PR #249, opened 2019-11-21
Filters moved to the store. Dynamic ListPage construction from the config json-file

OlegMakarenko:
- **Dropdown** options and current value moved to the store
- Added **PageList** component which will be used as a template
- Added config file `src/config/list-pages.json` for the router and PageList layout instructions

Resolved: #246


src/config/list-pages.json
```diff
@@ -0,0 +1,47 @@
+{
+    "pages": [
+        {
+            "path": "/blocks",
+            "name": "blocks",
+            "props": {
+              "namespace": "block",
```
AnthonyLaw: Just my opinion, show change the name "namespace" to "storeNamespace".
incase miss understand, `mosaic namespace`

src/router.js
```diff
@@ -18,78 +18,87 @@
...
       component: () =>
         import('./views/Home.vue')
     },
-    {
-      path: '/blocks',
-      name: 'blocks',
-      component: () =>
-        import('./views/Blocks.vue')
-    },
+    // {
```
AnthonyLaw: i think just remove the unused code.

src/store/transaction.js
```diff
@@ -79,15 +79,23 @@ export default {
...
+      height: el.blockHeight,
+      deadline: el.deadline,
+      transactionHash: el.transactionHash,
+      type: el.transactionBody.type,
+      fee: el.fee,
+      signer: el.signer,
+      recipient: el.transactionBody.recipient
     })),
+    // getTimelineFormatted: (state, getters) => getters.getTimeline.current.map(el => ({
+    //   height: el.height,
```
AnthonyLaw: i think should remove as well, if it's not longer using.

src/config/list-pages.json
OlegMakarenko: Good suggestion, agree

PR #281, opened 2019-12-24
Detail pages layout moved to config file

OlegMakarenko:


evias:
ah wait its breaking tests though

AnthonyLaw:
update the lint patch will fix travis issue.

evias:
what is the linter issue and why does the code not respect it?

PR #284, opened 2020-01-07
Delete netlify badge

dgarcia360:
Removing Netlify until we have Catapult https nodes.


AnthonyLaw:
> Removing Netlify until we have Catapult https nodes.

sure thing, we can do that later.

PR #287, opened 2020-01-13
Receipt implementation

AnthonyLaw:
### Added
- Receipt In block #260 
- Balance Change Receipt
- Balance Transfer Receipt
- Inflation Receipt
- ArtifactExpiryReceipt
- Address / Mosaic resolution statement


src/format.js
```diff
@@ -524,6 +524,91 @@ const formatMetadatas = metadatas => {
   }))
 }
 
+const formatReceiptStatements = receipts => {
+  let balanceChangeReceipt = []
+  let balanceTransferReceipt = []
+  let inflationReceipt = []
+  let artifactExpiryReceipt = []
+
+  receipts.map(receipt => {
```
OlegMakarenko: Since there is no assignment, I think it's can be better with forEach()

AnthonyLaw: sure, but what is mean no assignment?

Issue #223, opened 2019-11-12
Nodes Page uses Static Data


Alexhuszagh:
Nodes page doesn't show any real information about the nodes.

PR #289, opened 2020-01-14
Node list

AnthonyLaw:
### Added
- Node list page #223 
- showing roles, network, IP/DNS, Port, address and friendly name of the node


src/config/peers-api.json.mt
```diff
@@ -1,10 +1,11 @@
 {
-    "defaultNode": "http://api-harvest-20.us-west-1.nemtech.network:3000",
+    "defaultNode": "http://api-harvest-20.eu-west-1.nemtech.network:3000",
     "nodes": [
         "http://api-01.mt.us-west-2.nemtech.network:3000",
         "http://52.194.207.217:3000",
         "http://103.3.60.174:3000",
         "http://13.114.200.132:3000",
         "http://47.107.245.217:3000"
```
evias: please also update that defaults list to use testnet hosts :+1:

src/infrastructure/getNode.js
```diff
@@ -0,0 +1,37 @@
+/*
```
evias: change file name to `getNodePeers`

src/infrastructure/getNode.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ */
+
+import axios from 'axios'
+import http from './http'
+import dto from './dto'
+import format from '../format'
+
+class sdkNode {
+    static getNodes = async () => {
```
evias: getNodePeers

src/infrastructure/getNodePeers.js
```diff
@@ -22,14 +22,12 @@ import dto from './dto'
...
       const path = `/node/peers`
       const response = await axios.get(http.nodeUrl + path)
 
       const nodes = response.data.map(node => dto.createNodeInfoFromDTO(node, http.networkType))
 
-      console.log(format.formatNodesInfo(nodes))
-
       return format.formatNodesInfo(nodes)
     }
 }
```
evias: i dont think this has to be a class, can be just exporting the function: 

```
export getNodePeers = async () => {
    // ...
}
```

and then all imports just become :

```
import {getNodePeers} from '../infrastructure/getNodePeers'
```

OlegMakarenko: Since this class is totally static there is no sense in it. I agree with Greg it can be simple export of functions. But we have whole infrastructure in this static-class approach. I think it is make sense to keep everything in one style whether single functions or static-class.

What do you think @evias  @AnthonyLaw ?

AnthonyLaw: I think yes, we should keep everything in one style and using function approach.

Issue #292, opened 2020-01-20
Mosaic list: the network currency should always be in first position


decentraliser:
I think it is necessary to implement this feature, currently, nem.xem is always the last one

http://explorer.nemtech.network/account/TCORAO7MC7XH7ZN5H7RWZXRW5SBFB5DD6JPSMTDF
![image](https://user-images.githubusercontent.com/36499647/72695644-38002480-3b74-11ea-8116-d140d633eebe.png)

Issue #291, opened 2020-01-19
Searchbar could accept mosaicId and namespaceId


decentraliser:
Would be nice to be able to input namespaceId and mosaicId, as well as namespaceName maybe in the searchbar.

As URI such as below work, it shouldn't be too hard to implement

http://explorer.nemtech.network/mosaic/00B2A35D018CED36

AnthonyLaw:
@decentraliser thanks for bring this up.

Issue #252, opened 2019-11-22
Reduce the number of columns for the mobile view


OlegMakarenko:
![image](https://user-images.githubusercontent.com/33131259/69436411-fc023c00-0d49-11ea-8793-9b2391d6b0bd.png)

PR #295, opened 2020-01-21
Added MosaicId and NamespaceId support to the SearchBox. Network native mosaic first on the list. Reduced the number of columns for the mobile view

OlegMakarenko:
Resolved: #291 
Resolved: #292 
Resolved: #252 
Add names of the columns to config which have to be displayed in mobile view
`src\config\list-pages.json`
```
{
    "pages": [
        {
            "path": "/blocks",
            "name": "blocks",
            "props": {
                "storeNamespace": "block",
                "title": "blocks",
                "hasInfoText": true,
                "mobileColumns": [
                    "height",
                    "age",
                    "transactions",
                    "fee",
                    "harvester"
                ]
            }
        }
}

### CODE REVIEW

File: src/store/ui.js [hunk 1]
```diff
@@ -173,6 +201,9 @@ export default {
   }
 }
 
+const isMosaicId = (str) =>
```
AnthonyLaw: MosaicID and NamespaceID is same length actually. 
maybe we should rename it to `isMosaicOrNamespaceId`

AnthonyLaw:
for this method `isMosaicId`, it's only able to identify the mosaicID or namespaceID.

I like to suggest, if we can add Dropwdown to filter search Type 
- Account (Address, PublicKey)
- Transaction
- Mosaic (mosaicID or mosaic name)
- Namespace (namespaceID or namespace name)

src/store/ui.js
OlegMakarenko: Yes, good proposal, thanks. Changed in last commit

OlegMakarenko:
> for this method `isMosaicId`, it's only able to identify the mosaicID or namespaceID.
> 
> I like to suggest, if we can add Dropwdown to filter search Type
> 
> * Default (block height)
> * Account (Address, PublicKey)
> * Transaction
> * Mosaic (mosaicID or mosaic name)
> * Namespace (namespaceID or namespace name)
> 
> @evias @OlegMakarenko

What is the probability of a coincidence of MosaicId and NamespaceId?

Issue #311, opened 2020-02-04
Transaction paginator problem in block detail.


AnthonyLaw:
there is a bug with Transaction Paginator, it only permits to read 100 transactions in a block. 
Following block has 1500 transactions but pagination stops at 20. 

http://explorer-xym.nemtech.network/block/8953

PR #313, opened 2020-02-05
Block transaction list pagination fix. Transaction status widget

OlegMakarenko:
Resolved: #311


src/infrastructure/getTransaction.js
```diff
@@ -28,6 +28,27 @@ const QueryParams = nem.QueryParams // Travis patch
...
+          resolve(transactionStatus)
+        })
+        .catch(error => {
+          transactionStatus.message = error.response.data.message
+          transactionStatus.detail = error.response.data
+          resolve(transactionStatus)
+        })
+    })
+  }
+
```
AnthonyLaw: You can calling the status function here

```js
http.transaction.getTransactionStatus(hash).toPromise()
```

AnthonyLaw:
I thinking `Transaction Status` should direct display `Transaction Info`

- create 1 more `Card` in Transaction Info, if the Txs Status is `Error`.

So in the user experience,  transaction hash should insert in `Search bar`  rather than new StatusWidget.

src/infrastructure/getTransaction.js
OlegMakarenko: Thanks

PR #345, opened 2020-02-26
New Timeline, helper fetchFunction and Loading component changes

OlegMakarenko:
- Added new  **Timeline** class
- Added **Filter** class
- Added **DataSet** class
- Reorganized Store init/fetch functions
- Added `managers` support for DetailPage and ListPage
- Added `storeNamespaces` array prop to page config (init uninit)
Resolved: #333 

- Added **fetch function-wrapper**

Usage for ListPage store async functions:
```
await helper.fetchData(async () => { 
      commit("setData", await sdkNamespace.getNamespacesFromIdWithLimit(Constants.PageSize));
}, commit)
```

Is equal to:
```
commit('setLoading', true)
commit('setError', false)
      try {
                  commit("setData", await sdkNamespace.getNamespacesFromIdWithLimit(Constants.PageSize));
      } catch (e) {
                  console.error(e)
                  commit('setError', true)
      }
commit('setLoading', false)
```
- Added **Timeline pagination** for DetailPage and TableListView
```
...
"timelinePagination": true,
"timelineGetter": "block/blockTransactionsTimeline",                   
...
```

- Added mini prop for small view of Loading component. Used in TableListView for Pagination loading state
- Added `right` aligment prop to Dropdown


src/infrastructure/getBlock.js
```diff
@@ -33,27 +33,7 @@ class sdkBlock {
...
-    if (txList.length > 0) {
-      transactionId = txList[txList.length - 1].transactionId
-      const page = await this.getTransactionsByBlockHeight(blockHeight, transactionId)
-      txList = [...txList, ...page]
-    }
-    return format.formatBlocktransactions(txList)
-  }
-
-  static getBlockTransactionsFromId = async (blockHeight, transactionId) => {
+  static getBlockTransactionsFromIdSmall = async (blockHeight, transactionId) => {
```
AnthonyLaw: the name `Small` doesn't sound mean anything.
maybe just change the name `getBlockTransactions`

src/store/timeline2.js
```diff
@@ -0,0 +1,159 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import Constants from '../config/constants'
+
+export default class Timeline2 {
```
evias: what is the difference with the `Timeline` class and why not replace it?

Also,if not replacing  I would expect a more explanatory title than adding a 2

AnthonyLaw:
@OlegMakarenko you need to fix the lint issue, than i think go to go :)

PR #356, opened 2020-03-08
Rebranding

AnthonyLaw:
Rebase with master.


.gitignore
```diff
@@ -22,8 +22,7 @@ yarn-error.log*
 *.sw?
 
 # Custom config files.
-src/config/endpoints.json
-src/config/peers-api.json
+src/config/setup.json
```
evias: doesnt it use the peers-api.json file anymore? If not, we need to know how it changed because it will affect the build process (before automatic build, we copy the peers-api.json into the repo).

AnthonyLaw: peers-api.json and endpoints.json already merged into `setup.json`.

it wont affect the build  process, because all already predefine in `setup.json`.

but if you want make some change, you can edit `setup.json`

view more on template file [setup.json.mt](https://github.com/AnthonyLaw/nf-catapult-explorer/blob/rebranding/src/config/setup.json.mt)

evias: it will affect the jenkins build process ; not the package build

PR #394, opened 2020-04-10
Page assembler

OlegMakarenko:
1. Added **CardTable** component. Combine TableViews (list and info depend on data type), Dropdown filter, InfoText and pagination. Ready to use as a widget, just set a Data Manager getter 

Props:

- `title` - String - Text to show as a title
- `hasInfoText` - Boolean - Show text field (InfoText) in the card header
- `infoTextGetter` - String - InfoText getter name
- `hasFilter` - Boolean - Data manager has filter. Show Dropdown
- `managerGetter` - String - Data Manager getter 
- `dataGetter` - String - Data getter name. When use both with the `managerGetter`, data is fetching from `dataGetter` and error / loading state from `managerGetter`
- `pagination` - String ('disabled' | 'client' | 'server' | 'custom') - Pagination type
- `paginationOptions` - Object - Options for the 'custom' pagination
- `pageSize` - Number - Page size for 'client' pagination
- `columns` - Array - Columns to show in table list
- `mobileColumns` - Array - Columns to show in table list (mobile view)
- `errorMessage` - String - Custom error text

2. Added *PageAssembler view component. Now can be use to assemble list, detail, home, statistic pages. 

Page config:
```
"path": "/path", 
"name": "name", // page name
"meta": {
        "group": "list-page", // page group
        "keepAliveGoTo": ["detail-page"] // do not uninitialize when go to some of this groups
},
"props": {
        "layout": "flex", // layout type: flex, bootstrap or grid
        "storeNamespaces": ["account", "chain", "block", "transaction"], // init. and uninit. this stores
        "schema": [
            {
                "layoutOptions": "full-width", // use class names depends of layout type (ex. bootstrap - "col-xs-12 col-lg-6")
                "type": "BaseInfoWidget" // Component (widget) name. should be added to PageAsembler.components,
                "propA": true // Fill component props..
            },
          ] 
}
```
3. **TableView**. Changed names for some properties and methods

4. **Store/ui keyPages** moved to **config/key-redirects.json**

.


src/router.js
```diff
@@ -20,9 +20,20 @@ import Vue from 'vue'
 import Router from 'vue-router'
 import listPages from './config/list-pages'
 import detailPages from './config/detail-pages'
-
+import PageAssembler from './views/PageAssembler.vue'
+import pages from './config/pages'
+console.log(pages)
```
AnthonyLaw, line 25: should delete the console.log

src/components/containers/CardTable.vue
```diff
@@ -0,0 +1,198 @@
...
+      type: Object,
+      default: () => ({})
+    },
+    // client-side pagination pageSize
+    pageSize: {
+      type: Number,
+      default: 10
+    },
+    // columns to show in the table
+    columns: {
```
AnthonyLaw, line 106: Just suggestion, do you think is good idea rename to `fields` it sound more neutral.
`columns` shoud be fine for data list (Array) but for `Object {}` sound wrong.

Reason: I added fliter for Object data as well, i call it `fields`.

OlegMakarenko: Yes, it makes sense

src/router.js
OlegMakarenko: thanks

Issue #400, opened 2020-04-15
Refactor TableView 


OlegMakarenko:
1. Refactor TableView, TableListView, TableInfoView
2. Add fields components
3. Add Icons for transaction types
4. Fix Aggregate Transaction modal styles
5. Add Mosaics component

Issue #391, opened 2020-04-07
links on footer not balanced in layout


jabo38:
It would be nice if the links were centered and when the links switch from being only one line to two lines that it splits the links in half so that half of links are in one line and half are in the second. 

Same would apply to when it switches to three lines. So that the links would be divided equally by 1/3 or as close to as possible.  

Below is an image of "Faucet 2" being in a line by itself and being stuck to the right. Looks kind of odd. 
<img width="2069" alt="image" src="https://user-images.githubusercontent.com/10395907/78707043-afbd3000-78d5-11ea-95a2-d171c2e5e955.png">

And here is an image with "Faucet 1" and "Faucet 2" looking a bit odd in layout. 
<img width="1751" alt="image" src="https://user-images.githubusercontent.com/10395907/78707176-e2ffbf00-78d5-11ea-92f4-4a817ae9a328.png">

Ideally I would say when there isn't enough room in one row, then the second row would start of automatically with Github, Faucet 1, and Faucet 2 all being in the second line.

PR #409, opened 2020-04-23
TableView refactoring. Part 1

OlegMakarenko:
Resolved: #400

1. TableView refactored. Now uses **field** components for custom data types:

1.1. Added Icons for transaction types. **`<TransactionType />`** component

Transaction icons

1.2. Added inline mosaic list component **`<MosaicsField />`**

Modal + mosaics

1.3. Added **`<ArrayField />`** component. Displays arrays values. It also uses Truncate and clickable properties of TableView as simle table items

Multisig

2. Fixed DataManagers. Always use `uninitialize()` method when destroy a parent page. 

3. Changed color of Modal title

4. Resolved: #391


src/config/key-redirects.json
```diff
@@ -16,7 +16,7 @@
     "restrictionAddressDeletions": "account",
     "accountAliasName": "account",
     "cosignatories": "account",
-    "multisigAccounts": "account",
+    "multisigAccounts_": "account",
```
AnthonyLaw, line 19: it's typo for in the end of the word `_` ?

OlegMakarenko: It is item of. `multisigAccounts` array. I agree this is not intuitive. I will make some changes

PR #446, opened 2020-05-14
added transaction list for account and block

AnthonyLaw:


src/store/manager/Pagination.js
```diff
@@ -59,9 +59,17 @@ export default class Timeline {
   get index() {
     return (
             this?.pageInfo?.pageNumber
-              ? this.pageInfo.pageNumber - 1
+              ? this.pageInfo.pageNumber
               : void 0
-    ) || 0
+    ) || 1
+  }
```
OlegMakarenko, line 65: Can you please comment on this?

src/store/manager/Pagination.js
```diff
@@ -59,9 +59,17 @@ export default class Timeline {
...
+
+  get lastPage() {
+    return (
+            this?.pageInfo?.totalPages
+              ? this.pageInfo.totalPages
+              : void 0
+    ) || '..'
```
OlegMakarenko, line 72: We should return only a `number`. This '..' better to use in the ui component

PR #451, opened 2020-05-19
New filter approch for Pagination. Added icons for dropdown

OlegMakarenko:
Usage (Pagination constructor):
```
filter: [
     {
        label: <Dropdown option label>,
        icon: <icon class string>,
        value: {
          <searchCriteria key>: <searchCriteria value>
        }
      }
]
```

image

### DIFF
--- a/src/infrastructure/TransactionService.js
+++ b/src/infrastructure/TransactionService.js
@@ -176,8 +176,9 @@ class TransactionService {
       pageNumber,
       pageSize,
       orderBy: 'desc',
-      transactionTypes: filterVaule === '0' ? [] : [filterVaule],
-      group: 'Confirmed'
+      transactionTypes: [],
+      group: 'Confirmed',
+      ...filterVaule
     }
 
     const transactions = await this.searchTransactions(searchCriteria)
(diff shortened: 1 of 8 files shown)


src/store/transaction.js
```diff
@@ -35,10 +35,118 @@ const managers = [
     pageInfo: {
       pageSize: 20
     },
-    filter: {
-      0: 'All',
-      ...Constants.TransactionType
-    }
+    filter: [
```
AnthonyLaw, line 38: can we move this filter under config?
because we can reuse for detail-pages's card filter.

src/infrastructure/TransactionService.js
```diff
@@ -176,8 +176,9 @@ class TransactionService {
       pageNumber,
       pageSize,
       orderBy: 'desc',
-      transactionTypes: filterVaule === '0' ? [] : [filterVaule],
-      group: 'Confirmed'
+      transactionTypes: [],
+      group: 'Confirmed',
```
AnthonyLaw, line 180: i think you can remove 
`transactionTypes` and `group`
because i saw you already pass `transactionTypes` from store.

src/store/transaction.js
OlegMakarenko: Yes, good idea about config file. I will see what we can do

src/infrastructure/TransactionService.js
OlegMakarenko: It is default values. For example, when we pass some transactionTypes in the Store, we should not add `gruop: 'Confirmed'` to each option.

Issue #429, opened 2020-05-04
[Enhancement] Translation keys


AnthonyLaw:
- Some frequently use keyword
- Transaction type 
- Apply snake_case 
Apply across the all page.

AnthonyLaw:
snake_case may no a good options for explorer, because Service return data in camelCase and it's dynamic design.

so for the translation key we will remain in camelCase.

cc @evias

PR #486, opened 2020-06-02
Add translation

AnthonyLaw:
### Added
- Transaction type
- Error message

### Updated
- Rename key to camelCase

Resolved: #429


src/components/fields/TransactionType.vue
```diff
@@ -48,39 +49,59 @@ export default {
...
-                typeof type === 'string' &&
-                this.value
-                  .toUpperCase()
-                  .includes(
-                    type.toUpperCase()
-                  )
-      )
+    extractTransactionType(value) {
+      if (typeof this.value === 'string')
+        return Number(value.split('_').pop())
```
OlegMakarenko, line 104:
```
if (typeof value === 'string')
        return `Number(value.split('_').pop())
```

or

```
if (typeof this.value === 'string')
        return `Number(this.value.split('_').pop())
```

AnthonyLaw: thanks 👍

Issue #488, opened 2020-06-06
number and hash of transactions is incorrect (76436 instead 76435)


garm01:
Blocks first show the incorrect transaction and hash and after a while the correct number and hash of transactions is displayed

![Block76435](https://user-images.githubusercontent.com/46584115/83943567-73d31800-a7fd-11ea-9d81-050db51e11e0.png)
![image](https://user-images.githubusercontent.com/46584115/83943599-b7c61d00-a7fd-11ea-801d-1e0d6b9f1f46.png)

AnthonyLaw:
thanks for the feedback, I try it on my end, it seem everything is working fine.

![image](https://user-images.githubusercontent.com/5649156/83971713-7887f100-a90f-11ea-90f8-c6233809091f.png)

garm01:
please take a look at the new blocks now

garm01:
![20200607_17:10](https://user-images.githubusercontent.com/46584115/83972695-eddccd00-a8e1-11ea-99e5-cba3e508e1dd.png)

AnthonyLaw:
so i check on the problem, it may cause by Vuex issue.
seem the cache still remain the previous visited block transaction.

After refresh the page, all is seem working.

garm01:
I'm sorry, but I'm not sure if the bug is fixed:
![Screenshot_20200619](https://user-images.githubusercontent.com/46584115/85174899-9ee64e80-b276-11ea-9b18-15a399c9b6d8.png)

PR #490, opened 2020-06-08
to re-fetch timeline data into detail pages.

AnthonyLaw:
### Bug fix
- always re-fecth timeline data. Resolved: #488


src/store/manager/Timeline.js
```diff
@@ -65,7 +65,7 @@ export default class Timeline {
 
   initialFetch() {
     if (!this.initialized) {
-      this.initialized = true
```
OlegMakarenko, line 68: This line should be left to proper work of keep-alive

AnthonyLaw: yeah, but any way to force fetch the latest intead of remain the old data ?

Issue #330, opened 2020-02-14
AccountDetail. Add harvesting information 


evias:
Harvesting information should contain :

..TBD

AnthonyLaw:
Accounts Detail will show

- Total harvested blocks by that account
- harvested blocks list by that account

AnthonyLaw:
merged into Master

PR #513, opened 2020-07-03
Accout harvested info

AnthonyLaw:
### Added
- Total Harvested blocks Info into Account Detail
- Harvested block list into Account Detail

Resolved : #330


AnthonyLaw:
yeah, i will check on this, because i think keep the `master` as `0.9.5.1`
wait for 96x network is up, we can merge this to master branch.

PR #525, opened 2020-07-13
Transaction Graphic

OlegMakarenko:
- **GraphicComponent** - added **global properties** (TransactionGraphic scope)
- **GraphicComponent** - added click **handlers**
- Added **mobile view** support
- **New rules** for circle icons display

**New transaction** type **graphics** added:
1. AddressAliasTransaction (link & unlink)
2. MosaicAliasTransaction (link & unlink)
3. NamespaceRegistrationTransaction

**New components** added:
1. AddCircle
2. NamespaceCircle
3. NamespaceIcon
4. NamespaceListPopover
5. NamespaceUnlinkCircle
6. TablePopover

image

image


src/components/widgets/TransactionGraphicWidget.vue
```diff
@@ -0,0 +1,77 @@
...
+
+  data() {
+    return {
+      TransactionTypes: Constants.TransactionType
+    }
+  },
+
+  computed: {
+    isWidgetShown() {
+      return this.data.type === 16724 ||
```
AnthonyLaw, line 45: I thinking it will be good, 
directly import `transactionType` from `symbol-sdk` rather than hard coded the transaction type.

OlegMakarenko: The reason is to avoid using sdk inside the Graphic components. It may be an issue in components lib. But for the explorer should be fine. Changed

Issue #533, opened 2020-07-14
PageAssembler. Mobile view doesn't work properly 


OlegMakarenko:
`fields` property doesn't affect data to display in the mobile view
![Screenshot_2020-07-15-02-33-54-279_com android chrome](https://user-images.githubusercontent.com/33131259/87486475-0c9e5400-c644-11ea-9759-2e9c3044533c.png)

PR #544, opened 2020-07-16
Iss#533 fix PageAssembler mobile view table fields

OlegMakarenko:
Resolved: #533

### DIFF
--- a/src/config/pages/transaction-list.json
+++ b/src/config/pages/transaction-list.json
@@ -22,7 +22,6 @@
                     "deadline",
                     "transactionHash",
                     "transactionDescriptor",
-                    "maxFee",
                     "signer",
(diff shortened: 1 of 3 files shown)


src/config/pages/transaction-list.json
```diff
@@ -22,16 +22,14 @@
                     "deadline",
                     "transactionHash",
                     "transactionDescriptor",
-                    "maxFee",
```
AnthonyLaw, line 25: any reason remove the maxfee from the list?

OlegMakarenko: `maxFee` is not a criteria to find transaction in the list in most cases. In desktop view in my opinion it is better to show `value` instead

Issue #549, opened 2020-07-21
Internal `Transaction ID` is being treated and displayed as an entity attribute 


Vektrat:
We put a lot of insistence to this specific matter, and @martinayora just noticed that it is being disregarded anyway in this client (this is also the reason why the ID is out of the transaction context in the json object, unsure how SDK is handling/forwarding this).

`Transaction ID` is an internal attribute extremely dependent on the database layer that should die at the step where pagination is managed, yet it is being clearly brought to the user's eyes - giving the false impression of it being the transaction id, where in reality this concept does not exist - it is a database-transaction-entity-id (ObjectId), which has no place in the client domain.

For these reasons, and due to the distributed nature of the project, the ID of a transaction will change among different nodes, even for the same transaction, and cannot be trusted as a fixed reference for this object.

We believe this concept should be completely removed from the interface, or it will cause obvious confusion.

![image](https://user-images.githubusercontent.com/434196/88036250-95852600-cb43-11ea-8baa-d0a4c4b1cde7.png)

OlegMakarenko:
Thanks @Vektrat , we will fix it

PR #551, opened 2020-07-21
Remove transactionId

AnthonyLaw:
### Remove
- all the transactionId related data

Resolved: #549

### DIFF
--- a/e2e/pageDetail/transactions/accountAddressRestriction.spec.js
+++ b/e2e/pageDetail/transactions/accountAddressRestriction.spec.js
@@ -1,4 +1,5 @@
 import config from '../../config/network.conf.json'
+import datafiled from '../../config/datafiled.json'
 
 describe('Symbol Explorer Transaction detail page for Account Address Restriction.', () => {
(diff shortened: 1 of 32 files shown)


e2e/pageDetail/transactions/accountAddressRestriction.spec.js
```diff
@@ -15,7 +16,7 @@ describe('Symbol Explorer Transaction detail page for Account Address Restrictio
         })
 
         it('render correct transaction info titles', ()=> {
-            const items = ['Block Height', 'Transaction Hash', 'Transaction ID', 'Date', 'Deadline', 'Fee', 'Signature', 'Signer', 'Status', 'Confirmation']
+            const items = datafiled.transactionInfoField
```
OlegMakarenko, line 19: `transactionInfoField` maybe `transactionInfoFields` if there is an array?

AnthonyLaw: nice catch!

Issue #560, opened 2020-07-24
Changing node doesn't work


cryptoBeliever:
How changing node should work?

Right now no matter if I:
1) Change node (without refresh)
2) Change node and refresh 

Each time I see only request to other than selected node (http://api-01.us-west-1.096x.symboldev.network:3000 instead selected api-01.eu-central-1.096x.symboldev.network)

AnthonyLaw:
Which explorer are you face this issue?
I tried on http://explorer-96x.symboldev.network/
it seem working.

the node change, will do
- everytime you selected `endpoint`, it will fitler healtly Node from the node list endpoint.  `/node/health`
- if selected node is **Not** `health`, it will random pick the health endpoint to load the data.
- if selected node is  **Good**, it will set the selected endpoint as default, and load the data.

the default endpoint store in the Vue state layer, when you go to the `Next` or other action, it should load with with the new endpoint.

Once you refresh the page, the Vue state will refresh to be default endpoint.

cryptoBeliever:
Yes. I'm using http://explorer-96x.symboldev.network/.

Maybe some example.

0. I'm on page http://explorer-96x.symboldev.network/mosaics
1. I'm opening page and my default node is api-01.eu-central-1.096x.symboldev.network
2. I'm changing node to api-01.us-east-1.096x.symboldev.network
3. After refresh selected node goes back to api-01.eu-central-1.096x.symboldev.network but in network I see requests to http://api-01.us-west-1.096x.symboldev.network:3000/mosaics?pageSize=25&pageNumber=1&order=desc ...

AnthonyLaw:
>After page refresh (F5) selected node goes back to api-01.eu-central-1.096x.symboldev.network

Once you refresh the page, the Vue state will be reset.

I will look on that.
thanks :)

PR #566, opened 2020-07-30
Bug fix change node

AnthonyLaw:
### Added
- localstorage for node url

### Bug fixes
- reload detail page, after change the node url

Resolved: #560


src/store/index.js
```diff
@@ -91,6 +91,18 @@ export default new Vuex.Store({
...
+			case 'account-detail':
+				return helper.logError(dispatch, 'account/fetchAccountDetail', route.params);
+			case 'block-detail':
+				return helper.logError(dispatch, 'block/fetchBlockInfo', route.params);
+			case 'mosaic-detail':
+				return helper.logError(dispatch, 'mosaic/fetchMosaicInfo', route.params);
+			case 'namespace-detail':
+				return helper.logError(dispatch, 'namespace/fetchNamespaceInfo', route.params);
+			case 'transaction-detail':
+				return helper.logError(dispatch, 'transaction/getTransactionInfoByHash', route.params);
```
OlegMakarenko, line 105: What is a purpose of this changes?

AnthonyLaw: To re-fetch data from new endpoint, after change the node in all those detail pages

OlegMakarenko: PageAssembler page has `initAction` property of the action name which dispatches on page load. Does it work?

AnthonyLaw: It only work in the listing page, but not detail page.

OlegMakarenko: It means there is a bug. This actions should be dispatched via PageAssembler (initAction prop). I will approve the PR, because this is a topic of another issue.

OlegMakarenko (approved):

PR #637, opened 2020-09-28
added Docker auto build script

### LINKED ISSUE #636: Deployment. Auto Docker image build in Travis
- Build on Release Branch, and push in docker hub
- Build on Main Branch, and push in docker hub

AnthonyLaw:
Resolved #636


travis/docker.sh
```diff
@@ -0,0 +1,50 @@
...
+
+echo "Creating image ${DOCKER_IMAGE_NAME}:${CURRENT_VERSION}"
+docker build -t "${DOCKER_IMAGE_NAME}:${CURRENT_VERSION}" .
+
+if [ "$TRAVIS_BRANCH" = "$MAIN_BRANCH" ]
+then
+    echo "Building for ${$MAIN_BRANCH} branch..."
+    echo "Docker build latest version"
+    docker tag ${DOCKER_IMAGE_NAME}:${CURRENT_VERSION} ${DOCKER_IMAGE_NAME}:latest
+    docker push "${DOCKER_IMAGE_NAME}:latest"
```
fboucquez, line 34: `latest` may be confusing as it's an alpha, not an official release. People may use that instead of the `release` one. 

I would have 2 tags for alpha 

- ${CURRENT_VERSION}-alpha 
- "${CURRENT_VERSION}-alpha-$(date +%Y%m%d%H%M)"

The second one is for people that want a specific alpha. If there is a new alpha with the same version but you have the old one locally, Docker doesn't download it (without removing the local image). If the client specifies the latest timestamp it will be downloaded.

travis/docker.sh
```diff
@@ -0,0 +1,50 @@
...
+
+    echo "Docker build alpha version"
+    docker tag ${DOCKER_IMAGE_NAME}:${CURRENT_VERSION} ${DOCKER_IMAGE_NAME}:${VERSION}-alpha
+    docker push "${DOCKER_IMAGE_NAME}:${VERSION}-alpha"
```
fboucquez, line 38: where is VERSION defined? Should it be CURRENT_VERSION?

Both if are pretty similar, you can probably merge them together having the tag (alpha, release) name in variables. But this is up to you.

fboucquez (changes requested):

travis/docker.sh
AnthonyLaw: great, i look make more sense 👍

travis/docker.sh
AnthonyLaw: >where is VERSION defined? Should it be CURRENT_VERSION?

my mistaken, good catch!

I will update it.

fboucquez:
@AnthonyLaw , please double-check that Travis has all the env set. I would hide private ones like DOCKER_USERNAME and DOCKER_PASSWORD but keep visible DOCKER_IMAGE_NAME, MAIN_BRANCH, RELEASE_BRANCH for reference

AnthonyLaw:
@fboucquez yes, I added travis env. All is ready 👍

travis/docker.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [ "$TRAVIS_BRANCH" = "$RELEASE_BRANCH" ]
+then
+    echo "Building for ${$RELEASE_BRANCH} branch..."
+    echo "Docker tagging release version"
+    docker tag "${DOCKER_IMAGE_NAME}:${CURRENT_VERSION}" "${DOCKER_IMAGE_NAME}:release"
+
+    echo "Docker pushing release"
+    docker push "${DOCKER_IMAGE_NAME}:release"
+    docker push "${DOCKER_IMAGE_NAME}:${CURRENT_VERSION}"
```
fboucquez, line 50: Automatic steps you can do in the release branch:

* tag
* update patch version
* push the version update

Have a look at the release.sh we are using in many projects using the "release" branch trigger process.

https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/main/travis/release.sh

fboucquez:
One question @AnthonyLaw , is it possible to send env parameters or to mount a config file into the explorer image to tune the explorer configuration? If I want to change the rest endpoints or change the network type.

AnthonyLaw:
> One question @AnthonyLaw , is it possible to send env parameters or to mount a config file into the explorer docker image to tune the explorer configuration? If I want to change the rest endpoints or change the network type.

Nope, Explorer is statics web application, it's not support env parameters.
Every change need to prepare before it build.

travis/docker.sh
AnthonyLaw: Yeah, we have it https://github.com/nemgrouplimited/symbol-explorer/blob/main/travis/release.sh

But i seperate it, because I put everything releated to Docker in one file.

fboucquez:
Hi @AnthonyLaw sorry for the late response.

About the explorer configuration, I'll do a feature request. I would like to make the explorer configurable for different environments and network types so we can apply the configuration when booting an image. Explorer could define a default configuration if no custom is provided.

About the release.sh, I've missed that in the Travis file. You may want to swap, first do the docker image release push, and then the call release.sh that tags and upgrades the version.

travis/docker.sh
fboucquez: I've missed that in the Travis file. 

You may want to swap the steps in Travis YAML.  First, do the docker image release push, and then call release.sh that tags and increments the version.

AnthonyLaw:
> Hi @AnthonyLaw sorry for the late response.
> 
> About the explorer configuration, I'll do a feature request. I would like to make the explorer configurable for different environments and network types so we can apply the configuration when booting an image. Explorer could define a default configuration if no custom is provided.
> 
> I recommend burning a version by doing a dummy release. It's better to fail now that fail when we are doing a system-wide release. I know from experience!

Alright, I will take a look, how can we add environments variable in the statics web application after the build.
if you have any though, let me know as well.

fboucquez (approved):

fboucquez (approved):

Wayonb (approved):

OlegMakarenko (approved):

Issue #644, opened 2020-09-29
Rich list showing empty


Vektrat:
Not sure if this worked before, just noticed it, if it's supposed to work with XYM then the filtered ID is wrong ^^

Also note that you can customize the currency (not necessarily XYM), as a suggestion for an extra feature if not in the roadmap

AnthonyLaw:
@Vektrat Good catch, thanks mate :)

> Also note that you can customize the currency (not necessarily XYM), as a suggestion for an extra feature if not in the roadmap

Yeah, it can build `Advance Search` to fully utilize filter enpoint.

PR #647, opened 2020-09-29
bug fixes

AnthonyLaw:
### Added
- Fixes missing total txs number in block list.
- Fixes rich list filter.

Resolved: #644

### DIFF
--- a/src/config/filters.js
+++ b/src/config/filters.js
@@ -141,7 +141,7 @@ export const account = [
 		icon: 'mdi-circle',
 		value: {
 			orderBy: AccountOrderBy.Balance,
-			mosaicId: new MosaicId('5E62990DCAC5BE8A')
+			mosaicId: new MosaicId('5B66E76BECAD0860')
 		}
 	}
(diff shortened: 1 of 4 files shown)


src/config/filters.js
```diff
@@ -141,7 +141,7 @@ export const account = [
 		icon: 'mdi-circle',
 		value: {
 			orderBy: AccountOrderBy.Balance,
-			mosaicId: new MosaicId('5E62990DCAC5BE8A')
+			mosaicId: new MosaicId('5B66E76BECAD0860')
```
OlegMakarenko, line 144: Should be get from the config or network

OlegMakarenko (approved):

src/config/filters.js
AnthonyLaw: Yeah, but I try access from `http.networkCurrency.mosaicId`, but `filters.js` is execute before `http`.

OlegMakarenko (approved):

PR #685, opened 2020-10-13
Custom config.js that loads windows.globalConfig

fboucquez:
Hacky way of fixing  https://github.com/nemgrouplimited/symbol-explorer/issues/683

Atm, globalConfig is injected with vue at build time. This doesn't allow custom network configurations.

This fix uses a config.js file loaded in the index.html before the generated app.js starts and uses it.

Ideally, we should use ajax to load the config before/while Vue is bootstrapping. But the current code has the static `globalConfig` references. The globalConfig would be undefined before Vue can load it. In this case, the config.js is loaded before the whole app.


AnthonyLaw (approved):
Thanks a lot mate.

I'm thinking of using Axios fetch `setup.json` during App created() and save the data into the Vuex store, access it `setup.json` from getters.

for now, this should be good enough.

fboucquez:
Hi @AnthonyLaw, you are totally right. 

Using Axios to load the config.json when Vue is booting is the way to go. I hack my way in by loading a public config.js file before the application. I'm in a rush to get explorer up and running in bootstrap and I didn't have time to do it properly. 

Atm, the globalConfig is used statically, the explorer needs a little bit of refactoring so the different components load the config object from the Vue store or it's passed as a param. 

This PR does the job and it would be nice to have images configurable now, but if we can do it the right way, let's merge a better one.

AnthonyLaw:
> Hi @AnthonyLaw, you are totally right.
> 
> Using Axios to load the config.json when Vue is booting is the way to go. I hack my way in by loading a public config.js file before the application. I'm in a rush to get explorer up and running in bootstrap and I didn't have time to do it properly.
> 
> Atm, the globalConfig is used statically, the explorer needs a little bit of refactoring so the different components load the config object from the Vue store or it's passed as a param.
> 
> This PR does the job and it would be nice to have these configurable images automatically when merging to main. But if we can do it the right way, let's merge a better one.

No problem, let discuss in the upcoming short call and decide later 👍

AnthonyLaw (approved):

PR #690, opened 2020-10-15
add server

OlegMakarenko:
Resolved: #728 
1. Added server
 - to start `node server.js`
 - returns `ENV` merged with `default` config (`src/config/default.json`)
 - also returns explorer's static files
2. Changed client config loading
 - fetch config from the sever on the app start
 - then init Vue and other modules
 - if server returns empty config, use `src/config/default.json`


AnthonyLaw:
Here is my feedback.

1. The server should able to read `.env` from the root. Every time restarts the server. It should read the latest config from .env

2. Add `start` in the package's script to execute `node server.js.`

3. Update runner in Dockerfile.
```
...
FROM node:10-alpine AS runner
WORKDIR /app
COPY --from=builder /app /app
EXPOSE 4000
CMD ["npm", "start"]
```

So when creating the docker container with Explorer image, we can always pass different config from outside.

OlegMakarenko:
@AnthonyLaw 
1. it reads defaulr.json and replace properties with ENV variables if such provided
2. added
3. added
4. added ability to set up the server port from ENV

AnthonyLaw (approved):

Issue #694, opened 2020-10-19
Use prettier format code.


AnthonyLaw:

cryptoBeliever:
@AnthonyLaw about what this issue is?

AnthonyLaw:
yes, we are now using eslint rules from `Symbol` submodule.
prettier won't be required.

PR #696, opened 2020-10-19
Fix style

AnthonyLaw:
### Added
- prettier code format library
- added `npm run style:fix` to fix lint.

Resolved: #694 

**Note**: prettier dropped support for Node versions older than 10
https://prettier.io/blog/2020/03/21/2.0.0.html


OlegMakarenko (commented):
Does the `<templete>` section of .vue now has 2 space indents. The `<script>` and `<style>` sections have tabs I guess. I'm not sure it is a good idea to mix it

OlegMakarenko (commented):
This PR changes the **tabs** indet to **spaces**. The cons form me that we will be no longer able to configure own comfortable tabsize in the editor. For example in Github you can specify the tab size adding `?ts=4` to the url. What do you think?

AnthonyLaw:
> This PR changes the **tabs** indet to **spaces**. The cons form me that we will be no longer able to configure own comfortable tabsize in the editor. For example in Github you can specify the tab size adding `?ts=4` to the url. What do you think?

I believe it still able to configure the tab size in your editor, just need to execute the command `npm run style` prettier will update the style.
we can discuss it later :)

AnthonyLaw:
redo is needed.

Issue #675, opened 2020-10-08
Finalized Blocks are not up to date


garm01:
Finalized Blocks are 4-5 hours behind

![Block_Finalized not up to date](https://user-images.githubusercontent.com/46584115/95494478-46aa9800-099e-11eb-8b36-544ca0f8b717.png)

AnthonyLaw:
it depends on the network nodes configurationion.

http://api-01.ap-southeast-1.0.10.0.x.symboldev.network:3000/chain/info
http://api-01.eu-central-1.0.10.0.x.symboldev.network:3000/chain/info

Base on the API endpoint response.

```
{
"scoreHigh": "0",
"scoreLow": "604448414600593350",
"height": "38750",
"latestFinalizedBlock": {
    "finalizationEpoch": 54,
    "finalizationPoint": 45,
    "height": "38160",
    "hash": "398574CDDE8CB62D8BE47311CDEB37BE314CEE7350F803629566D027F7E8EAE1"
  }
}
```

Current height with Finalized block different in 590 blocks (avg 30 sec per blocks).
behind 4 - 5 hours is valid.

AnthonyLaw:
thanks for @garm01 bring out this problem.

The finalized block height is always far behind to voting node.
I think we should always use a voting node in Explorer.

what do you think about this? @Wayonb

Wayonb:
@AnthonyLaw Are you suggesting only add voting nodes to explorer?  I think both view are good to have.  Maybe, we might want to show users which node are voting or not so they can pick the correct end point.

AnthonyLaw:
> @AnthonyLaw Are you suggesting only add voting nodes to explorer? 

Yes, because the user can always get the latest finalized block height. (to avoid confusion) 

>I think both view are good to have. Maybe, we might want to show users which node are voting or not so they can pick the correct end point.

Good point, the node domain name can the `voting` prefix.

AnthonyLaw:
We will add extra node information (such as Voting node) into the Home page Base Info widget.

PR #723, opened 2020-11-16
Display node type

AnthonyLaw:
Resolved: #675 

image


src/config/i18n/en-us.json
```diff
@@ -356,5 +356,6 @@
     "finalizationPoint": "Finalization Point",
     "finalizedHash": "Finalized Hash",
     "nodeDetailError": "Failed to fetch node detail.",
-    "extendGraphicValue": "Info"
+    "extendGraphicValue": "Info",
+    "votingNode": "(Voting)"
```
OlegMakarenko, line 360: I think we should add more detail. Somithing like this:

`Voting node. Finalized height may be behind` (or ahead)

OlegMakarenko: And `( )` I think should be hardcoded in the component

OlegMakarenko (approved):

OlegMakarenko (approved):

Issue #796, opened 2021-02-25
Node Rewards Widget. Handle the no rounds state


OlegMakarenko:
**CB**
The widget shows the "failed to parse" error message when there is no test results in the response. 

**EB**
Display the "The node is not included in the round yet" message instead.

OlegMakarenko:
done

Issue #795, opened 2021-02-25
UI home page issues


OlegMakarenko:
1. 
![image](https://user-images.githubusercontent.com/33131259/109156056-0f842a00-7779-11eb-8dbf-6b658bc1a45d.png)
2. 
![image](https://user-images.githubusercontent.com/33131259/109156814-f7f97100-7779-11eb-9688-182ed9093242.png)

3. Base Info -> **Block Height** should be **Chain Height**

4. Missing loading spinner in the Base Info

5. Add nodes total count

OlegMakarenko:
done

Issue #721, opened 2020-11-12
Disable XEM price graph.


AnthonyLaw:
- apply it during Symbol launch.
- if we do have XYM graph to show, we will disable the price graph, until the chat is ready.

AnthonyLaw:
It may be related to #720 
if we disable XEM price graph from homepage.
we need found something to replace.

OlegMakarenko:
done

PR #799, opened 2021-02-25
Fixes

OlegMakarenko:
Resolved: #795 
Resolved: #796 
Resolved: #721

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "symbol-explorer",
-  "version": "0.9.0",
+  "version": "0.9.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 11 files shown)


AnthonyLaw:
I think issue #721 should put into separate fix PR.

Because once the PR merge, we can't prepare any release until launch I guess.

OlegMakarenko:
@AnthonyLaw I think it should be merged and tested before launch, so we will have a time to fix if any issue occurs

AnthonyLaw:
> @AnthonyLaw I think it should be merged and tested before launch, so we will have a time to fix if any issue occurs

yes, for sure.
just that as long as we no more release before launch, all is fine :)

AnthonyLaw (approved):

PR #832, opened 2021-03-17
Searchbox. Show error when user enters nis1 address

OlegMakarenko:
image


src/config/default.json
```diff
@@ -10,6 +10,13 @@
 			"http://api-01.us-west-1.testnet.symboldev.network:3000"
 		]
 	},
+	"nis1Nodes": [
```
AnthonyLaw, line 13: Any reason we need this?

src/config/i18n/en-us.json
```diff
@@ -1,5 +1,7 @@
 {
     "langName": "en-us",
+    "errorNisAddressNotAllowed": "You are trying to search the NEM NIS1 address in the Symbol Explorer. Please use NEM Explorer instead",
+    "errorNothingFound": "Nothing found..",
```
segfaultxavi, line 4:
```suggestion
    "errorNisAddressNotAllowed": "You entered a NEM NIS1 address in the Symbol Explorer. Please use NEM Explorer instead.",
    "errorNothingFound": "Nothing found.",
```
Could the link to the NEM Explorer be given?

segfaultxavi (changes requested):

src/config/default.json
OlegMakarenko: Changed solution a bit. Node list removed

src/config/i18n/en-us.json
OlegMakarenko: Thank you, translation improved.

> Could the link to the NEM Explorer be given?
This is a good idea but we don't have an official explorer for NIS1

AnthonyLaw (approved):

PR #852, opened 2021-03-25
mobile view fix

### LINKED ISSUE #829: When viewing  tx “show details” on mobile no way to exit details.
When I’m on mobile (both Android and iOS) and I go to this tx http://explorer.symbolblockchain.io/transactions/6BBA746B6C3C6502F8F20DF2369E78193FFBDA67CF506260480644903E67B796

and I go to show details of each aggregate to see the messages, I can’t close out of the message. I have to leave the app. 

9DFE3B64-1289-4EB4-B462-30662C63DF78

AnthonyLaw:
Resolved #829 

Screenshot 2021-03-25 at 6 16 13 PM

### DIFF
--- a/src/components/containers/Modal.vue
+++ b/src/components/containers/Modal.vue
@@ -11,7 +11,9 @@
 		header-text-variant="title"
 	>
 		<template v-slot:default>
-			<slot name="body" />
+			<div class="mobile-model-content">
+				<slot name="body" />
+			</div>
 		</template>
 	</b-modal>
 </template>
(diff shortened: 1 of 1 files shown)


src/components/containers/Modal.vue
```diff
@@ -30,3 +32,11 @@ export default {
 	}
 };
 </script>
+
+<style lang="scss" scoped>
+@media screen and (max-width: 420px){
+	.mobile-model-content {
+		height: 300px;
```
OlegMakarenko, line 39: I think `max-height: 75%` should be more flexible

AnthonyLaw: yeah, but % is not working in this case.

OlegMakarenko (approved):

Issue #864, opened 2021-04-02
Failed to fetch transaction list in Partial filter


AnthonyLaw:

AnthonyLaw:
merged

PR #865, opened 2021-04-02
Fix partial transaction list

AnthonyLaw:
- Fixed partial transaction in Transaction List
- Fixed  partial transaction in Account Transaction List

image

Resolved: #864


OlegMakarenko (approved):
I think we should move all transaction logic to the `TransactionService`

Issue #856, opened 2021-03-29
Add new data field in node list.


AnthonyLaw:
- Rest server version `/node/server`

AnthonyLaw:
merged

PR #866, opened 2021-04-05
Add rest version

AnthonyLaw:
### Added
- software version component
- showing rest server version on node list

Resolved: #856 

image

### DIFF
--- a/src/config/i18n/en-us.json
+++ b/src/config/i18n/en-us.json
@@ -388,5 +388,7 @@
 	"viewMoreStatistics": "View More Statistics",
     "nodeCountChartTitle": "Node count over time",
     "timestamp": "Timestamp",
-    "symbolTime": "Symbol Time"
+    "symbolTime": "Symbol Time",
+    "restVersion": "Rest Version",
+    "softwareVersion": "Version"
 }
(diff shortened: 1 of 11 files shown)


src/config/i18n/en-us.json
```diff
@@ -388,5 +388,7 @@
 	"viewMoreStatistics": "View More Statistics",
     "nodeCountChartTitle": "Node count over time",
     "timestamp": "Timestamp",
-    "symbolTime": "Symbol Time"
+    "symbolTime": "Symbol Time",
+    "restVersion": "Rest Version",
```
segfaultxavi, line 392: Is the user familiar with the concept of a REST API? Should we talk about REST at all, or maybe it is enough to say API?

In any case, REST should be capitalized.
```suggestion
    "restVersion": "REST Version",
```

segfaultxavi (changes requested):

src/config/i18n/en-us.json
AnthonyLaw: I think this is more for the Developer to view.
IMO REST should be fine because the GitHub repo called `catapult-rest`

OlegMakarenko (approved):

Issue #867, opened 2021-04-05
Harvesting fees not presented for pretty address in url


cryptoBeliever:
Harvesting fees are presented only if we put address without dashes.

Example:
http://explorer.symbolblockchain.io/accounts/NDKYBCN3GMJ6TM6TLVWLUYVXIS7WTZ4A2QXJ5HI
http://explorer.symbolblockchain.io/accounts/NDKYBC-N3GMJ6-TM6TLV-WLUYVX-IS7WTZ-4A2QXJ-5HI

For the second one harvesting blocks section is not presented:

![image](https://user-images.githubusercontent.com/17684281/113620497-abe3fb00-965a-11eb-9a76-bc3b1dffb929.png)

AnthonyLaw:
merged into Dev

PR #868, opened 2021-04-06
fixed pretty address in url not able retrieve data.

AnthonyLaw:
Resolved: #867

### DIFF
--- a/src/store/account.js
+++ b/src/store/account.js
@@ -180,26 +180,32 @@ export default {
 			context.getters.timeline.setStore(context).initialFetch();
 		},
 
-		// Fetch data from the SDK By Address.
+		// Fetch data from the SDK By plain Address.
 		async fetchAccountDetail(context, payload) {
-			if (!helper.isAccountAddress(payload.address))
-				payload.address = await helper.decodeToAddress(payload.address);
+			let address = payload.address;
+
+			try {
+				address = Address.createFromRawAddress(address).plain();
+			}
+			catch (e) {
+				address = await helper.decodeToAddress(address);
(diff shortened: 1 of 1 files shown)


src/store/account.js
```diff
@@ -180,26 +180,32 @@ export default {
 			context.getters.timeline.setStore(context).initialFetch();
 		},
 
-		// Fetch data from the SDK By Address.
+		// Fetch data from the SDK By plain Address.
 		async fetchAccountDetail(context, payload) {
-			if (!helper.isAccountAddress(payload.address))
-				payload.address = await helper.decodeToAddress(payload.address);
```
OlegMakarenko, line 186: IMO the flow should be

- check if the value is plain address, then return it
- remove the hyphens and check if the value is plain address, then return it
- check if the value is publicKey, then convert to address and return it
- check (fetch) if the value is an address alias then return address

src/store/account.js
```diff
@@ -180,26 +180,32 @@ export default {
 			context.getters.timeline.setStore(context).initialFetch();
 		},
 
-		// Fetch data from the SDK By Address.
+		// Fetch data from the SDK By plain Address.
```
OlegMakarenko, line 183: Should be `Fetch account data from the SDK by address, publicKey or alias namespaceName

src/store/account.js
AnthonyLaw: The first check should be address or alias, right?

OlegMakarenko (dismissed):

OlegMakarenko (approved):

Issue #846, opened 2021-03-22
Add Export feature in Explorer nodes list.


AnthonyLaw:
Add Export list into dropdown.
can be export in csv format.

![image](https://user-images.githubusercontent.com/5649156/111966705-2d536d80-8b32-11eb-9549-24e41f383907.png)

AnthonyLaw:
merged into Dev

PR #870, opened 2021-04-08
Export csv

AnthonyLaw:
### Added
- Export button component
- Export CSV in Node list.

image

Screenshot 2021-04-09 at 12 47 27 AM

Resolved: #846

### DIFF
--- a/src/components/controls/DropdownFilter.vue
+++ b/src/components/controls/DropdownFilter.vue
@@ -5,19 +5,31 @@
 		:size="_size"
 		:right="right"
 	>
-		<b-dropdown-item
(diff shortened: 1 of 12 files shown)


OlegMakarenko (approved):
Looks good. But it will not work with the array of filters. In this case the scv button should be assigned to the table, not to the dropdown selector.


image

AnthonyLaw:
> Looks good. But it will not work with the array of filters. In this case the scv button should be assigned to the table, not to the dropdown selector.
> 
> image

Agree, we should follow that design in the wallet.

Issue #873, opened 2021-04-12
block creation time is displayed incorrectly for a short time.


garm01:
When a new block appears in the explorer, " 364 days ago " is often displayed for a (very)short time under "Created". 

---

![blockcreation](https://user-images.githubusercontent.com/46584115/114457124-f70f8800-9bdd-11eb-8193-9e774b6b889a.png)

OlegMakarenko:
Merged into `dev`

PR #877, opened 2021-04-20
Bug fix

AnthonyLaw:
# Bug fixes
- prevent time compared in negative value and cause the problem #873
- removed `math.floor` in TransactionPerBlock statistics chart data.

image
 
 no sure why I can't add you as Reviewers, so I just tag here @segfaultxavi  ;)


segfaultxavi:
That average looks much nicer 😄

OlegMakarenko (approved):

PR #884, opened 2021-05-10
Added the Node Rewards Enrollment views

OlegMakarenko:
@AnthonyLaw @yilmazbahadir

1. Added Enrollment **List page**
2. Added Enrollment **filter**
3. Added Enrollment **Detail page**
4. Added **Enrollment table** to the Account Detail page
5. Added **button** to the Node List page to access [1]
6. Fixed the last page pagination issue

### Access from the **Node List** page:
image
image
image

### Access from the **Node Details** page:
image
image


src/infrastructure/NodeService.js
```diff
@@ -342,6 +342,45 @@ class NodeService {
...
+
+		if (signerPublicKey)
+			url += `/signerPublicKey/${signerPublicKey}`;
+
+		const response = (await Axios.get(url, { params: searchCriteria })).data;
+
+		const data = {
+			...response,
+			...response.pagination,
+			isLastPage: response.data.length < pageSize
```
AnthonyLaw, line 365: That is a condition if that is the last page.
and assume `pageSize` is 20, and `response.data.length` exactly 20.

so `isLastPage` should be `true` , but `response.data.length < pageSize` return `false` am I right?

AnthonyLaw (dismissed):
Look good 👍🏼

src/infrastructure/NodeService.js
OlegMakarenko: Yes, you are right. I've also added a check to the Pagination class. So if reqested page is empty, it will stay on previous page.
Otherwise we need solution which we used before - make a double reqest (pageNumber, pageNumber+1)

PR #899, opened 2021-06-01
feat: add support of the EarlyAdoption and Ecosystem NR payouts

OlegMakarenko:
- Added the EarlyAdoption and Ecosystem payout support to the NodeService and UI changes
- Updated changelog
- Updated symbol-sdk version

image


src/infrastructure/NodeService.js
```diff
@@ -311,13 +311,26 @@ class NodeService {
...
+			let route = '/payouts';
+
+			switch (rewardProgram) {
+			case 'EarlyAdoption':
+				route = '/payouts/earlyadoption';
+				break;
+			case 'Ecosystem':
+				route = '/payouts/ecosystem';
+				break;
+			case 'SuperNode':
```
AnthonyLaw, line 329: Do we still need this case `SuperNode`? seem it already handle by default.

OlegMakarenko: For better code understanding :)

AnthonyLaw (approved):

Issue #927, opened 2021-08-17
3001 HTTPS support


fboucquez:
Atm I can see that there is HTTP and 3000 hardcoded when resolving the gateway URL address of a node. 

https://github.com/symbol/symbol-explorer/blob/dev/src/infrastructure/NodeService.js#L117

The idea is to convert all our nodes (and community eventually) to https port 3001. Is it possible to update the code? Ideally, the URL resolution should "ping" 3001 and 3000 to know if the node has HTTPS support yet.

Note that this is different from the explorer configuration where we should be able to update without a code change.

https://github.com/symbol/symbol-explorer/blob/dev/src/config/default.json#L5

We should revisit the statistics service for a similar gateway URL resolution from a node's hostname

fboucquez:
About the exception handling, you can see the problem by going to this HTTP node:

https://explorer.rockbear.io/nodes/965C7555721D1503F5024E02F42979FD6833142A63483CE3AB5E35F7EB094432

![image](https://user-images.githubusercontent.com/5390558/129761218-dd89b41c-7580-4350-9b18-834e51ea70a3.png)

AnthonyLaw:
> if a HTTPS explorer tries to talk to a HTTP node, the browser would reject the call.

not sure it's a good idea, because even nodes without HTTPS enabled, still consider valid nodes.

rg911:
would be better to show both http and https nodes, can probably add a `protocol` column next to the port in node list?

fboucquez:
Yes, we should show both http and https nodes. The problem would be that when asking a HTTP only node for the health and info, the browser will reject. HTTPS explorer cannot talk to a HTTP node for the state. See the screenshot above

AnthonyLaw:
> Yes, we should show both http and https nodes. The problem would be that when asking a HTTP only node for the health and info, the browser will reject. HTTPS explorer cannot talk to a HTTP node for the state. See the screenshot above

alright, got it now 👍🏼

AnthonyLaw:
> would be better to show both http and https nodes, can probably add a `protocol` column next to the port in node list?

yeah, but I think it should be test and store into the static service.

PR #932, opened 2021-09-08
Enable https

AnthonyLaw:
### Fix
- change ws -> wss.
- replace http -> https

### Add
- node https support verify

Resolved: #927


src/helper.js
```diff
@@ -125,11 +125,11 @@ class helper {
 			return url;
 	}
 
-	static httpToWsUrl(str) {
+	static httpToWssUrl(str) {
 		let url = new Url(str);
 
 		if (this.validURL(url)) {
-			url.set('protocol', 'ws:');
```
rg911, line 132: will it still support non https nodes?

src/infrastructure/NodeService.js
```diff
@@ -195,6 +195,10 @@ class NodeService {
             formattedNode.rolesRaw === 6 ||
             formattedNode.rolesRaw === 7
     	) {
+    		// If https not working, it will replace to http with port 3000
+    		if (!(await this.isHttpsEnabled(formattedNode.apiEndpoint)))
+    			formattedNode.apiEndpoint = formattedNode.apiEndpoint.replace('https', 'http').replace(':3001', ':3000');
```
rg911, line 200: what about calling the HTTPS URL by default, if the exception is thrown, try HTTP, rather than always makes an extra call?

src/config/default.json
```diff
@@ -1,20 +1,20 @@
...
+		"defaultNode": "https://ngl-dual-001.testnet.symboldev.network:3000",
 		"nodes": [
-			"http://ngl-dual-001.testnet.symboldev.network:3000",
-			"http://ngl-dual-101.testnet.symboldev.network:3000",
-			"http://ngl-dual-201.testnet.symboldev.network:3000",
-			"http://ngl-dual-301.testnet.symboldev.network:3000",
-			"http://ngl-dual-401.testnet.symboldev.network:3000",
-			"http://ngl-dual-501.testnet.symboldev.network:3000",
-			"http://ngl-dual-601.testnet.symboldev.network:3000"
+			"https://ngl-dual-001.testnet.symboldev.network:3000",
```
fboucquez, line 5: port will be 3001

src/infrastructure/NodeService.js
```diff
@@ -114,7 +114,7 @@ class NodeService {
             nodeInfo.roles === 3 ||
             nodeInfo.roles === 6 ||
             nodeInfo.roles === 7
-            	? 'http://' + nodeInfo.host + ':' + (globalConfig.apiNodePort || 3000)
+            	? 'https://' + nodeInfo.host + ':' + (globalConfig.apiNodePort || 3001)
```
fboucquez, line 117: you may need to 'ping' the community node to know if they have migrated to HTTPS 3001 or they are still in HTTP 3000

src/infrastructure/NodeService.js
fboucquez: If you are trying a URL, remember to add a timeout. If the port is close, the request may hang

src/infrastructure/NodeService.js
```diff
@@ -414,6 +418,16 @@ class NodeService {
 
 		return helper.convertArrayToCSV(formattedData);
 	}
+
+	static isHttpsEnabled = async (apiEndpoint) => {
+		try {
+			await Axios.get(apiEndpoint + '/node/info', { timeout: 3000 });
```
fboucquez, line 424: why axiox? Fetch is the library used in the sdk. You can add timeouts (not as simple but there is a way)

src/helper.js
AnthonyLaw: No, the default node should all be in HTTPS, right?
But let me just made a port checking, so it can switch between `wss` or `ws`.

src/infrastructure/NodeService.js
AnthonyLaw: Yes, I make a timeout `3000` in Axios call. it's good enough?

src/config/default.json
AnthonyLaw: good catch 👍🏼

src/infrastructure/NodeService.js
AnthonyLaw: There was few service use Axios as well such as node reward endpoint and statistic service endpoint.

src/infrastructure/NodeService.js
AnthonyLaw: Yes, this is formatting the default data when it returns for the statistic service or node peers.
I assume most of the nodes will be enabled `https`.

It will ping the node when the user visits `node/<publicKey>` in explorer.

Issue #902, opened 2021-06-01
Brightness levels of the display for the explorer.


ghost:
The explorer should offer the possibility of displaying different brightness levels ( **light, dimmed**, dark mode ).

PR #958, opened 2021-11-10
Dark mode

AnthonyLaw:
- Add toggle theme (light / dark) on menu
- Default theme will be dark
- Replace the transaction type icon
- Replace the new symbol icon
- Replace font type
- minor adjustment on Header and Footer

Resolved: #902


src/components/ThemeToggle.vue
```diff
@@ -0,0 +1,101 @@
...
+</template>
+
+<script>
+import IconMoon from '../styles/img/moon.png';
+import IconSun from '../styles/img/sun.png';
+
+export default {
+	mounted() {
+		// get theme from localstorage
+		let theme = localStorage.getItem('theme');
```
OlegMakarenko, line 35: I think we should not access localStorage directly from UI components. I would move it to store or service

src/components/menu/MobileMenu.vue
```diff
@@ -89,14 +93,25 @@ export default {
...
 		getNameByKey(e) {
 			return this.$store.getters['ui/getNameByKey'](e);
+		},
+
+		iconUrl(icon) {
+			switch (icon) {
+			case 'IconStatistics':
+				return this.IconStatistics;
+			case 'IconMosaics':
+				return this.IconMosaics;
```
OlegMakarenko, line 103: Why there is only `Statistic` and `Mosaic` icons?

src/components/menu/NavigationMenu.vue
```diff
@@ -68,33 +70,36 @@ export default {
...
 	methods: {
 		getNameByKey(e) {
 			return this.$store.getters['ui/getNameByKey'](e);
+		},
+		iconUrl(icon) {
+			switch (icon) {
+			case 'IconStatistics':
+				return this.IconStatistics;
+			case 'IconMosaics':
+				return this.IconMosaics;
```
OlegMakarenko, line 88: Why there is only `Statistic` and `Mosaic` icons?

src/components/ThemeToggle.vue
```diff
@@ -0,0 +1,101 @@
...
+
+			// Save theme in local storage
+			localStorage.setItem('theme', darkMode);
+		}
+	}
+};
+</script>
+
+<style lang="scss" scoped>
+.toogleTheme {
```
OlegMakarenko, line 70:
```suggestion
.toogle-theme {
```

And above

src/components/menu/MobileMenu.vue
AnthonyLaw: That 2 new icons from the design team.

src/components/ThemeToggle.vue
AnthonyLaw: The reason put into localstorage is to prevent theme reset.

OlegMakarenko (approved):

PR #980, opened 2021-11-24
Add mosaic flag on the mosics list

AnthonyLaw:
The flag column will show icons if the flag is true.

image


OlegMakarenko (dismissed):

src/infrastructure/MosaicService.js
```diff
@@ -123,7 +123,13 @@ class MosaicService {
    		data: mosaicInfos.data.map((mosaic) => ({
    			...mosaic,
    			ownerAddress: mosaic.address,
-   			mosaicAliasNames: this.extractMosaicNamespace(mosaic, mosaicNames)
+   			mosaicAliasNames: this.extractMosaicNamespace(mosaic, mosaicNames),
+   			mosaicFlags: {
+   				mutable: mosaic.supplyMutable,
```
OlegMakarenko, line 128: Should we keep property name `supplyMutable` consistent?

```suggestion
   				supplyMutable: mosaic.supplyMutable,
```

OlegMakarenko (dismissed):
@AnthonyLaw do we need translations for that?

AnthonyLaw:
> @AnthonyLaw do we need translations for that?

I think we can do it later.

OlegMakarenko (approved):

PR #1004, opened 2021-12-10
Add recent namespaces and mosaics in home page

AnthonyLaw:
### Add
- recent list widget components, can generate the widget dynamic base on the params.

Example
```
{
	title: 'namespaces',   // Recent component title
	listingPage: '/namespaces',   // route path
	dataGetter: 'namespace/getRecentList',   // access data from store
	header: 'namespaceName',   // Clickable header for the row
	fields: ['approximateExpired', 'ownerAddress']   // field below header
},
```

Screenshot 2021-12-11 at 4 18 36 AM


src/components/widgets/RecentListWidget.vue
```diff
@@ -0,0 +1,175 @@
+<template>
+	<Card :loading="loading">
+		<template #title>
+			{{getNameByKey('recent')}} {{getNameByKey(title)}}
```
OlegMakarenko, line 4: There may be a problem with some languages (eg. word order). I would use single key for translation.
```suggestion
			{{getNameByKey(title)}}
```

src/views/Home.vue
```diff
@@ -47,7 +59,29 @@ export default {
...
+		RecentTransactionsWidget,
+		RecentListWidget
+	},
+
+	data() {
+		return {
+			recentListWidgetSetup: [
+				{
+					title: 'namespaces',
+					listingPage: '/namespaces',
```
OlegMakarenko, line 71: Changing this field to `viewMoreLinkURL` could add more sense, what do you think?

src/components/widgets/RecentListWidget.vue
```diff
@@ -0,0 +1,175 @@
...
+		ButtonMore
+	},
+
+	computed: {
+		recentList() {
+			return this.$store.getters[this.dataGetter] || [];
+		},
+
+		loading() {
+			return !this.recentList.length;
```
OlegMakarenko, line 114: Why don't we use `dataManager.isLoading`?

src/components/widgets/RecentListWidget.vue
```diff
@@ -0,0 +1,175 @@
...
+			]
+		};
+	},
+
+	components: {
+		Card,
+		ButtonMore
+	},
+
+	computed: {
```
OlegMakarenko, line 108: Should we handle the error state?

src/components/widgets/RecentListWidget.vue
```diff
@@ -0,0 +1,175 @@
...
+		fields: {
+			type: Array,
+			require: true
+		}
+	},
+
+	data() {
+		return {
+			clickableKeys: [
+				'ownerAddress'
```
OlegMakarenko, line 98: Why is there only one `clickableKey`? Should we reuse the list from the `TableView` component?

src/views/Home.vue
```diff
@@ -47,7 +59,29 @@ export default {
...
+					dataGetter: 'namespace/getRecentList',
+					header: 'namespaceName',
```
OlegMakarenko, line 73: Changing this field to `headerField` could add more sense, what do you think?

OlegMakarenko (changes requested):

src/components/widgets/RecentListWidget.vue
AnthonyLaw: Yea, because I was thinking, it's a widget, not sure it's good to extend the `TableView` component.

src/components/widgets/RecentListWidget.vue
AnthonyLaw: it's just array list, didn't apply data manager

src/components/widgets/RecentListWidget.vue
AnthonyLaw: I don't think it need because it's just an array from the list, default is an empty array

OlegMakarenko (approved):

OlegMakarenko (approved):

PR #1008, opened 2021-12-14
Gh1005 node icon

AnthonyLaw:
## Add 
- Node role type icon on the node list

image

WIP - it need to change the node icon on the map


segfaultxavi:
Minor nitpick, the case of the tooltips is not consistent. I would change them to:

peer --> Peer
api --> API
VOTING --> Voting

PR #1013, opened 2021-12-17
Improvement blocks listing page

AnthonyLaw:


OlegMakarenko:
> add inflation.json

@AnthonyLaw
Is it the same for all networks? Should we consider returning this data from REST?

AnthonyLaw:
> Is it the same for all networks? Should we consider returning this data from REST?

@Wayonb what do you think about this add inflation data on REST?

Wayonb:
> @Wayonb what do you think about this add inflation data on REST?

Yes, inflation data can be different per network and would be better coming from REST.
I have looked the PR in detail but why are you hardcoding the inflation?

AnthonyLaw:
> > @Wayonb what do you think about this add inflation data on REST?
> 
> Yes, inflation data can be different per network and would be better coming from REST. I havent looked at the PR in detail but why are you hardcoding the inflation?

Reason for hardcore, because I think the value is fixed.
right now in Rest is only able to retrieve the single inflation rate from each block, using transaction statement and filter with receipt type `20803`

it may not well option to fire 10 requests to get an inflation rate from 10 blocks.

src/config/inflation.json
```diff
@@ -0,0 +1,425 @@
+{
```
Jaguar0625, line 1: why do we need to duplicate this instead of just loading an ini file copied from the catapult-client?

src/helper.js
```diff
@@ -697,6 +697,15 @@ class helper {
 	static getStartListIndex = (pageNumber, pageSize) => {
 		return pageNumber === 1 ? 0 : (pageNumber - 1) * pageSize;
 	}
+
+	static getInflationByHeight = (height) => {
+		const blocksHeight = Object.keys(inflation);
+
+		for (let i = 0; i < blocksHeight.length; i++) {
```
Jaguar0625, line 704: 1. `length - 1` or else last iteration will try to index length because of `i + 1`.
2. please use `++i` instead of `i++` whenever possible; former can generally be better optimized (although compilers/interpreters should be smart enough nowadays)
3. you should refer last value if height is past the last entry

Jaguar0625 (changes requested):

Wayonb:
> right now in Rest is only able to retrieve the single inflation rate from each block, using transaction statement and filter with receipt type 20803

Cann't you use the ``fromHeight`` and ``toHeight`` filter options so something like this?
https://001-joey-dual.symboltest.net:3001/statements/transaction?receiptType=20803&fromHeight=10000&toHeight=10010

AnthonyLaw:
> https://001-joey-dual.symboltest.net:3001/statements/transaction?receiptType=20803&fromHeight=10000&toHeight=10010

oh, I miss out on this endpoint. thanks for remind 👍🏼

src/config/inflation.json
AnthonyLaw: I was missing out on the `receipt` endpoint and was able to return a list of inflation.
`inflation.json` will be removed.

> just loading an ini file copied from the catapult-client
how does it work?
where can we find **ini** file?

src/helper.js
AnthonyLaw: In this method, I will remove and use the receipt endpoint.
Btw that is good feedback for me to improve the next method.

src/config/inflation.json
Jaguar0625: https://github.com/symbol/symbol-node-configurator/blob/main/templates/resources/config-inflation.properties

Jaguar0625:
please remember to add some regression tests for this change

Jaguar0625 (approved):

PR #1017, opened 2022-01-04
fix: only display network currency XYM

AnthonyLaw:
### Current behavior
- Owned mosaics will always sort XYM to the top, the `Account balance widget` will always display the first mosaic from the owned mosaics.

### What was the issue?
- If owned mosaics don't include XYM, the `Account Balance Widget` will be showing other mosaics. Which is confuses users.

### What's the fix?
- `Account Balance Widget` will only display in XYM.
- It will show 0 XYM if owned mosaics don't include XYM.
- Added Unit test.
- Test case
```
If the account has non-mosaic, it should display XYM::0
visit address TA2CXCBJH44ZS4Z3KCVG3KJHBP4FCW3U7B4OWJA

If the account has mosaics with XYM, it should display XYM with the amount.
visit address TBFKFNFAYN3WKQDG6QZMKCMS6K5QMVYQKIYL5UA

If the account has mosaics and is without XYM, it should display XYM::0
visit address TC7FDHJZ53SK5OFUIL2C6X25KQYBXHBCCWCSJGA
```

### DIFF
--- a/src/components/widgets/AccountBalanceWidget.vue
+++ b/src/components/widgets/AccountBalanceWidget.vue
@@ -75,7 +75,11 @@ export default {
 		},
 
 		balance() {
-			return this.data.mosaic?.amount || '0';
+			// Only display network Currency
+			if (this.data.mosaic && http.networkCurrency.mosaicId === this.data.mosaic.mosaicId)
+				return this.data.mosaic.amount;
+
+			return '0';
 		},
 
 		address() {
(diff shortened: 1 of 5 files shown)


src/components/widgets/AccountBalanceWidget.vue
```diff
@@ -89,16 +93,8 @@ export default {
...
-				const mosaicNamespaces = mosaicAlias.split('.');
-				const mosaicLastSubnamespace = mosaicNamespaces.pop();
-
-				return mosaicLastSubnamespace;
-			}
-
-			return this.data.mosaic?.mosaicId || http.networkCurrency.namespaceName.split('.').pop();
+			// Only display for network currency
+			return http.networkCurrency.namespaceName.split('.').pop();
 		},
```
Jaguar0625, line 98: is this widget intended to only ever displaying currency now?

if so, why the check in balance()?

AnthonyLaw: I'm not sure I understand your question correctly.

> is this widget intended to only ever displaying currency now?
> if so, why the check in balance()?

yes, It was displayed the first mosaic from an array, XYM will always sort to the first one.
If the account holding XYM, that is not a problem, because it showed XYM and balance.

If the account does not hold any XYM, the other mosaic will be displayed.

To check balance is ensure only display XYM rather than other mosaics.

Jaguar0625: i mean is AccountBalanceWidget intended to only ever show XYM balance or could it be used to show balance of other mosaics as well?

AnthonyLaw: yes only show XYM balance.

Jaguar0625 (approved):
👍

PR #1044, opened 2022-04-12
[explorer] fix: filter rich list

AnthonyLaw:
## What was the issue?
- `filters.js` loaded before the app requested the network to collect mosaic Id, so it caused `http.networkCurrency.mosaicId` to return undefined.
- it affected accounts pages `rich` list not working.

## What's the fix?
- let mosaicId to an empty value, if the value is empty, will assign a network mosaic id.

PS: I don't think this is a good solution, but still finding a way to load dynamic mosaic id into `filters.js`

### DIFF
--- a/src/config/filters.js
+++ b/src/config/filters.js
@@ -1,5 +1,4 @@
 import Constants from '../config/constants';
-import http from '../infrastructure/http';
 import {
 	TransactionType,
 	TransactionGroup,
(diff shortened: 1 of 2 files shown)


Jaguar0625:
wasn't this working before? did something change?

AnthonyLaw:
> wasn't this working before? did something change?

Before was working, mosaic's Id was hardcoded in config.
I believe it was affected by this PR https://github.com/symbol/explorer/pull/1010

Do you think it's possible to unit-test for this to ensure mosaic id is not undefined, I'm not sure how to do it.

Jaguar0625 (approved):

PR #1068, opened 2022-06-01
[explorer] perf: reduce block request to obtain timestamp

AnthonyLaw:
## What was the issue?
- Every transaction information called extra request from block endpoint to obtain timestamp.

## What's the fix?
- updated SDK to 2.0.1
- removed extra requests to obtain block timestamp and transaction effectiveFee.


Jaguar0625 (approved):
nice job 👍

Issue #1051, opened 2022-05-06
Invalid amount of transferred XYM presented on list and details when XYM sent both using namespace alias and using mosaic id


cryptoBeliever:
Description:
When adding XYM mosaic two times to mosaic array:
- once using mosaic id
- once using namespace alias
explorer presents only one value as transferred XYM. It should present the sum of it.

Example transfer:
https://symbol.fyi/transactions/03E52F8B2B3FF8598ABBB5C508FA0DABE18A767533B27779ADEC59FE9661E6B1

Transferred amount on:
- the lists
- transfer amount tooltip
is invalid. It presents the first entry but it should be the sum of both: 0.000005 + 0.000006 = 0.000011 XYM

PR #1069, opened 2022-06-02
[explorer] bug: fix display mosaic amount issue

AnthonyLaw:
## What was the issue?
- Mosaic amount wrong display, when sending transaction attached mosaic's Id and namespace. [read here ](https://github.com/symbol/explorer/issues/1051)
- Explorer filtered by mosaic id and showing the amount, instead of sum the amount balance.

## What's the fix?
- removed unused method `networkCurrencyMosaicBuilder`.
- sum the amount of mosaic filtered by mosaic id.
- minor refactor on helper method.

### DIFF
--- a/__tests__/helper.spec.js
+++ b/__tests__/helper.spec.js
@@ -11,7 +11,7 @@ const mockTestMosaic = {
 	namespaceName: 'namespace_mosaic'
 };
 
-describe.only('Helper', () => {
+describe('Helper', () => {
 	describe('hslToRgb should', () => {
 		it('return rgb', () => {
 			// Arrange:
@@ -69,7 +69,7 @@ describe.only('Helper', () => {
 		});
 	});
 
-	describe('mosaicsFieldObjectBuilder should', () => {
+	describe('mosaicsFieldObjectBuilder', () => {
 		beforeEach(() => {
 			stub(MosaicService, 'getMosaics').resolves(Promise.resolve([
 				TestHelper.mockMosaicInfo(mockTestMosaic.idHex, 'TC46AZWUIZYZ2WVGLVEZYNZHSIFAD3AFDPUJMEA', 10, 0)
@@ -99,17 +99,36 @@ describe.only('Helper', () => {
 			expect(result).toStrictEqual([
 				{
 					amount: '0.000001',
-					id: new MosaicId(http.networkCurrency.mosaicId).id,
-					mosaicAliasName: http.networkCurrency.namespaceName,
+					rawAmount: UInt64.fromUint(1),
 					mosaicId: http.networkCurrency.mosaicId,
-					rawAmount: UInt64.fromUint(1)
+					mosaicAliasName: http.networkCurrency.namespaceName
 				},
 				{
-					id: new MosaicId(mockTestMosaic.idHex).id,
 					amount: '4',
 					rawAmount: UInt64.fromUint(4),
 					mosaicId: mockTestMosaic.idHex,
-					mosaicAliasName: [mockTestMosaic.namespaceName]
+					mosaicAliasName: mockTestMosaic.namespaceName
+				}
+			]);
+		});
+
+		it('returns mosaic object when only network currency', async () => {
+			// Arrange:
+			const mockMosaics = [
+				new Mosaic(new MosaicId(http.networkCurrency.mosaicId), UInt64.fromUint(10)),
+				new Mosaic(new NamespaceId(http.networkCurrency.namespaceName), UInt64.fromUint(20))
+			];
+
+			// Act:
+			const result = await Helper.mosaicsFieldObjectBuilder(mockMosaics);
+
+			// Assert:
+			expect(result).toStrictEqual([
+				{
+					amount: '0.000030',
+					rawAmount: UInt64.fromUint(30),
+					mosaicId: http.networkCurrency.mosaicId,
+					mosaicAliasName: http.networkCurrency.namespaceName
 				}
 			]);
 		});
@@ -129,24 +148,41 @@ describe.only('Helper', () => {
 			// Assert:
 			expect(result).toStrictEqual([
 				{
-					id: new MosaicId(http.networkCurrency.mosaicId).id,
 					amount: '0.000003',
-					mosaicAliasName: http.networkCurrency.namespaceName,
+					rawAmount: UInt64.fromUint(3),
 					mosaicId: http.networkCurrency.mosaicId,
-					rawAmount: UInt64.fromUint(3)
+					mosaicAliasName: http.networkCurrency.namespaceName
 				},
 				{
-					id: new MosaicId(mockTestMosaic.idHex).id,
 					amount: '6',
 					rawAmount: UInt64.fromUint(6),
 					mosaicId: mockTestMosaic.idHex,
-					mosaicAliasName: [mockTestMosaic.namespaceName]
+					mosaicAliasName: mockTestMosaic.namespaceName
 				}
 			]);
 		});
+
+		it('returns empty array when mosaics not exist', async () => {
+			// Arrange:
+			const mockMosaics = [];
+
+			// Act:
+			const result = await Helper.mosaicsFieldObjectBuilder(mockMosaics);
+
+			// Assert:
+			expect(result).toStrictEqual(mockMosaics);
+		});
 	});
 
-	describe('resolveMosaicId should', () => {
+	describe('resolveMosaicId', () => {
+		let stubGetLinkedMosaicId = {};
+
+		beforeEach(() => {
+			stubGetLinkedMosaicId = stub(NamespaceService, 'getLinkedMosaicId').resolves(Promise.resolve(new MosaicId(mockTestMosaic.idHex)));
+		});
+
+		afterEach(restore);
+
 		it('returns id given MosaicId', async () => {
 			// Arrange:
 			const mockMosaicId = new MosaicId(mockTestMosaic.idHex);
@@ -155,7 +191,8 @@ describe.only('Helper', () => {
 			const result = await Helper.resolveMosaicId(mockMosaicId);
 
 			// Assert:
-			expect(result).toStrictEqual(mockMosaicId.id);
+			expect(result).toStrictEqual(mockMosaicId);
+			expect(stubGetLinkedMosaicId.callCount).toBe(0);
 		});
 
 		it('returns network mosaic id given network NamespaceId', async () => {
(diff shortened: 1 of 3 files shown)


src/helper.js
```diff
@@ -537,41 +539,35 @@ class helper {
...
-			throw new Error('Mosaic id does not match network Currency.');
+			if (idHex === http.networkCurrency.mosaicId) {
+				mosaicsFieldObject.push({
+					...mosaics[0],
+					rawAmount: UInt64.fromUint(sumAmount),
+					mosaicId: mosaics[0].id.toHex(),
+					amount: this.formatMosaicAmountWithDivisibility(sumAmount, http.networkCurrency.divisibility),
+					mosaicAliasName: http.networkCurrency.namespaceName
+				});
+			} else {
```
Jaguar0625, line 557: why do you need special casing for network currency? why doesn't normal resolution logic work?

AnthonyLaw: I excluded this because network currency info doesn't need a query from the network, so we can reduce 1 request for that (better perf).

In the `else` case, we requested mosaic info from the network.

Jaguar0625: that makes sense, but why `sumAmount` in `if` case but `mosaics[0].amount` in `else case?

AnthonyLaw: Good catch, it should in both.
I did some refactoring, and an extra unit test, to ensure those are covered.

src/helper.js
```diff
@@ -537,41 +542,37 @@ class helper {
...
-			rawAmount: mosaic.amount,
-			mosaicId: mosaic.id.toHex(),
-			amount: this.formatMosaicAmountWithDivisibility(mosaic.amount, http.networkCurrency.divisibility),
-			mosaicAliasName: http.networkCurrency.namespaceName
-		};
+				mosaicsFieldObject.push({
+					...mosaicField,
+					amount: helper.formatMosaicAmountWithDivisibility(sumAmount, divisibility),
+					mosaicAliasName: MosaicService.extractMosaicNamespace({ mosaicId: mosaics[0].id.toHex() }, mosaicNames)
+				});
```
Jaguar0625, line 571: you can go one step further refactoring. only remaining difference is divisibility used and derivation of mosaicAliasName. you can store those in local variables and then have only single push after if/else

__tests__/helper.spec.js
```diff
@@ -1,18 +1,167 @@
...
+			// Act:
+			const result = await Helper.resolveMosaicId(mockNamespaceId);
+
+			// Assert:
+			expect(result).toStrictEqual(new MosaicId(http.networkCurrency.mosaicId).id);
+		});
+	});
+
+	describe('getNetworkCurrencyBalance should', () => {
+		it('returns network currency balance', async () => {
```
Jaguar0625, line 131: add test where UNAVAILABLE is returned

__tests__/helper.spec.js
```diff
@@ -1,18 +1,167 @@
...
+					amount: '6',
+					rawAmount: UInt64.fromUint(6),
+					mosaicId: mockTestMosaic.idHex,
+					mosaicAliasName: [mockTestMosaic.namespaceName]
+				}
+			]);
+		});
+	});
+
+	describe('resolveMosaicId should', () => {
```
Jaguar0625, line 106: add test when mosaic being looked up is network currency mosaic

if possible, check if `NamespaceService.getLinkedMosaicId` mocked function is called or not

__tests__/helper.spec.js
```diff
@@ -1,18 +1,167 @@
...
+import { stub, restore } from 'sinon';
+import { Mosaic, MosaicId, UInt64, NamespaceId, NamespaceName } from 'symbol-sdk';
 
-describe('Helper', () => {
+const mockTestMosaic = {
+	idHex: '6B2E9EAF2632AEC8',
+	namespaceName: 'namespace_mosaic'
+};
+
+describe.only('Helper', () => {
```
Jaguar0625, line 13: drop .only

__tests__/helper.spec.js
```diff
@@ -1,18 +1,167 @@
...
+			// Act:
 			const rgb = Helper.hslToRgb(hue, saturation, lightness);
 
-			//Assert:
+			// Assert:
 			expect(rgb).toEqual({R: 215, G: 158, B: 250});
 		});
 	});
+
+	describe('mosaicsFieldObjectBuilder should', () => {
```
Jaguar0625, line 29: in describe, drop "should"
missing test where mosaics is empty

src/helper.js
```diff
@@ -537,41 +542,37 @@ class helper {
...
-						...resolvedMosaic,
-						rawAmount: resolvedMosaic.amount,
-						mosaicId: resolvedMosaic.id.toHex(),
-						amount: helper.formatMosaicAmountWithDivisibility(resolvedMosaic.amount, divisibility),
-						mosaicAliasName: MosaicService.extractMosaicNamespace({ mosaicId: resolvedMosaic.id.toHex() }, mosaicNames)
-					});
-				}
-			}
-		}
+		const uniqueMosaicIds = [...new Set(resolvedMosaics.map(mosaic => mosaic.id.toHex()))];
```
Jaguar0625, line 545: in line 525, why do you need to create new MosaicId, what is returned type of resolvedMosaic?

```
			const mosaicId = new MosaicId(resolvedMosaic.toHex()).id;
```

Jaguar0625: in line 530, why are you filtering out network mosaic id?

Jaguar0625: if you don't filter it out, uniqueMosaicIds would just be `[...new Set(resolvedMosaicIds)]`

__tests__/helper.spec.js
```diff
@@ -1,18 +1,167 @@
...
+					mosaicId: http.networkCurrency.mosaicId,
+					rawAmount: UInt64.fromUint(1)
+				},
+				{
+					id: new MosaicId(mockTestMosaic.idHex).id,
+					amount: '4',
+					rawAmount: UInt64.fromUint(4),
+					mosaicId: mockTestMosaic.idHex,
+					mosaicAliasName: [mockTestMosaic.namespaceName]
+				}
```
Jaguar0625, line 70: for readability, please use same property order for asserts of same object type

__tests__/helper.spec.js
```diff
@@ -1,18 +1,167 @@
...
+					mosaicAliasName: http.networkCurrency.namespaceName,
+					mosaicId: http.networkCurrency.mosaicId,
+					rawAmount: UInt64.fromUint(1)
+				},
+				{
+					id: new MosaicId(mockTestMosaic.idHex).id,
+					amount: '4',
+					rawAmount: UInt64.fromUint(4),
+					mosaicId: mockTestMosaic.idHex,
+					mosaicAliasName: [mockTestMosaic.namespaceName]
```
Jaguar0625, line 69: why is this array in this object but not previous one ?

src/helper.js
Jaguar0625: try to add test where 0 == resolvedMosaicIds.length

src/helper.js
```diff
@@ -537,41 +542,37 @@ class helper {
...
-	static networkCurrencyMosaicBuilder = mosaic => {
-		if (!(mosaic instanceof Mosaic))
-			throw new Error('It required Mosaic instance.');
+			const sumAmount = mosaics.reduce((acc, cur) => acc + Number(cur.amount.toString()), 0);
 
-		if (mosaic.id.toHex() !== http.networkCurrency.mosaicId)
-			throw new Error('Mosaic id does not match network Currency.');
+			const mosaicField = {
+				...mosaics[0],
+				rawAmount: UInt64.fromUint(sumAmount),
```
Jaguar0625, line 554: if there is precision issue, it's too late to convert to UInt64 here, and you probably want to use BigInt for arithmetic

src/helper.js
```diff
@@ -537,41 +542,37 @@ class helper {
...
+				Object.assign(mosaicField, {
+					amount: this.formatMosaicAmountWithDivisibility(sumAmount, divisibility),
+					mosaicAliasName: MosaicService.extractMosaicNamespace({ mosaicId: mosaics[0].id.toHex() }, mosaicNames)
+				});
+			}
+
+			mosaicsFieldObject.push(mosaicField);
+		});
+
+		return mosaicsFieldObject;
```
Jaguar0625, line 575: i think you can use map instead of forEach?

src/helper.js
```diff
@@ -245,14 +248,13 @@ class helper {
...
 	static getNetworkCurrencyBalance = mosaics => {
-		let mosaic = mosaics.find(mosaic =>
-			mosaic.id.toHex() === http.networkCurrency.mosaicId ||
+		const networkCurrencyMosaic = mosaics.filter(mosaic => mosaic.id.toHex() === http.networkCurrency.mosaicId ||
 			(mosaic.id instanceof NamespaceId &&
-				mosaic.id.toHex() === http.networkCurrency.namespaceId));
+			mosaic.id.toHex() === http.networkCurrency.namespaceId));
 
-		let balance = mosaic !== undefined ? this.toNetworkCurrency(mosaic.amount) : Constants.Message.UNAVAILABLE;
+		const totalNetworkCurrencyAmount = networkCurrencyMosaic.reduce((acc, cur) => acc + Number(cur.amount.toString()), 0);
```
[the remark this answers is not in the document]
AnthonyLaw: amount type is UInt64

src/helper.js
```diff
@@ -245,14 +248,13 @@ class helper {
 	 * @returns {string} network currency balance.
 	 */
 	static getNetworkCurrencyBalance = mosaics => {
-		let mosaic = mosaics.find(mosaic =>
-			mosaic.id.toHex() === http.networkCurrency.mosaicId ||
+		const networkCurrencyMosaic = mosaics.filter(mosaic => mosaic.id.toHex() === http.networkCurrency.mosaicId ||
```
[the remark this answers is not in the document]
AnthonyLaw: in the transaction, users are able to attach `UnresolvedMosaicId` (namespace id OR mosaic id)

src/helper.js
AnthonyLaw: That makes sense, I can just return rather than build and push to array.

src/helper.js
AnthonyLaw: > in line 525, why do you need to create a new MosaicId, what is returned type of resolvedMosaic?

it returns as `Id`, I will do some refactoring on this.

> in line 530, why are you filtering out network mosaic id?

those filtered mosaic ids need to call the API request mosaic info and namespace name, excluding network mosaic id, because we already have the information.

src/helper.js
AnthonyLaw: alright, I will take a look at this. 

`amount` from SDK is returned as `UInt64`, `rawAmount` will need to be UInt64 as well.

so we should convert `Uint64` -> `bigInt` and do arithmetic? after that convert back to `Uint64`?

__tests__/helper.spec.js
AnthonyLaw: so mosaic can have more than one alias name, this is the reason I use it as Array.

the previous one is network currency, which is that is not possible to have more than one alias name.

but it looks strange, I should convert it to the array.

src/helper.js
Jaguar0625: to make my point clearer, here is an example:

when you're doing arithmetic on Uint64s in JS, you want to do something like this [or equivalent]
> BigInt(Number.MAX_SAFE_INTEGER) + 2n - BigInt(Number.MAX_SAFE_INTEGER)
> 2

but you're effectively doing arithmetic with Number not BigInt, which could lead to wrong result:
> BigInt(Number.MAX_SAFE_INTEGER + 2) - BigInt(Number.MAX_SAFE_INTEGER)
> 1

__tests__/helper.spec.js
```diff
@@ -119,15 +119,15 @@ describe('Helper', () => {
 			]);
 		});
 
-		it('returns empty array when mosaics not exist', async () => {
+		it('returns empty array when mosaics does not exist', async () => {
```
Jaguar0625, line 122: does => do (because mosaics is plural)

Jaguar0625 (approved):

AnthonyLaw:
@Jaguar0625 I rebased it to top of dev

Jaguar0625 (approved):

PR #1070, opened 2022-06-03
[explorer] task: change timezone in local as default

AnthonyLaw:
## What was the issue?
- Current date is in UTC format, we like to change it to local date.

## What's the fix?
- `convertTimestampToDate` method to convert timestamp to date, it accepts `UTC` or `Local`.  default as `Local`
- removed unused `convertDeadlinetoDate` and `convertToUTCDate`

### DIFF
--- a/__tests__/components/fileds/DateField.spec.js
+++ b/__tests__/components/fileds/DateField.spec.js
@@ -21,7 +21,7 @@ const setupStoreMount = () => {
 
 	const propsData = {
 		timestamp: 1615853185,
-		isShowTimestamp: true
+		shouldShowTimestamp: true
 	};
 
 	return shallowMount(DateField, {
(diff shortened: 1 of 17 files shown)


AnthonyLaw:
image

@0x6861746366574  Should we keep the Expiration date in UTC format? IMO we should keep it in UT format, everyone will see the same expiration date.

__tests__/helper.spec.js
```diff
@@ -15,4 +16,32 @@ describe('Helper', () => {
 			expect(rgb).toEqual({R: 215, G: 158, B: 250});
 		});
 	});
+
+	describe('convertTimestampToDate should', () => {
```
Jaguar0625, line 20: ~~should~~ and change int to action verbs; i.e. convert => converts

__tests__/helper.spec.js
```diff
@@ -15,4 +16,32 @@ describe('Helper', () => {
...
+		});
+
+		it('convert timestamp in utc date', () => {
+			// Act:
+			const date = Helper.convertTimestampToDate(networkTimestamp, 'UTC');
+
+			// Assert:
+			expect(date).toEqual('2021-03-16 00:06:25');
+		});
+	});
```
Jaguar0625, line 46: consider adding test for unknown label too

Jaguar0625: alternatively, consider using enum instead of string for discriminator

__tests__/components/fileds/DateField.spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+	const store = new Vuex.Store({
+		modules: {
+			ui:uiModule
+		}
+	});
+
+	const propsData = {
+		timestamp: 1615853185,
+		isShowTimestamp: true
```
Jaguar0625, line 24: is => should

__tests__/components/fileds/DateField.spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+	it('renders date and timestamp', async () => {
+		// Act:
+		await wrapper.vm.showDate(timestamp);
+
+		// Assert:
+		expect(wrapper.html()).toMatch('<div class=\"date-container\">'
+            + '<span title=\"2021-03-16 08:06:25\">2021-03-16 08:06:25</span> '
+            + '<span class=\"timestamp\">Symbol Time : 1615853185</span></div>');
+	});
```
Jaguar0625, line 66: mixed tabs and spaces

__tests__/components/fileds/DateField.spec.js
```diff
@@ -0,0 +1,67 @@
...
+		restore();
+		unregister();
+	});
+
+	it('processes props data', ()=> {
+		// Assert:
+		expect(wrapper.vm.timestamp).toBe(timestamp);
+		expect(wrapper.vm.isShowTimestamp).toBe(true);
+	});
+
```
Jaguar0625, line 57: unsure what this test is adding

Jaguar0625: is there any case you can think of where this test would fail but following test would not?
in other words, isn't following test already implicitly testing that everything is passed down correctly?

AnthonyLaw: yeah, agree, I got it now.
I will add 1 more test case with `shouldShowTimestamp : false`

PR #1107, opened 2022-09-06
[explorer] bug: unable to display big list of aggregation transaction

AnthonyLaw:
## What was the issue?
- Unable to load the significant list aggregation transaction, when there are too many API requests it will have 429 errors.
- The explorer will always request mosaic info and namespace for every transaction which included mosaic. It works well in a single transaction, but when coming to aggregation transactions it will have performance issues because it included a list of the inner transactions.

## What's the fix?
- Instead of calling request every transaction, it will collect all mosaic from the inner transaction, and only make a single request to gather mosaic and namespace info. It will reduce calling API requests, and have better performance.
- added new method `getTransactionMosaicInfoAndNamespace` to request mosaic info and namespace.
- added unit test
- removed request mosaic info and namespace logic inside `mosaicsFieldObjectBuilder`


__tests__/infrastructure/CreateTransaction.spec.js
```diff
@@ -113,4 +119,164 @@ describe('CreateTransaction should', () => {
 			namespaceName: [namespace]
 		});
 	});
+
+	it('returns transferTransaction', async () =>  {
```
Jaguar0625, line 123: while this is true statment, test is really verifying that address and mosaics within transferTransaction are resolved, so something like 

resolves mosaics and addresses in transfer transaction 

would be more descriptive around intent of test imo

__tests__/infrastructure/TransactionService.spec.js
```diff
@@ -15,7 +15,7 @@ describe('Transaction Service', () => {
 				}
 			};
 
-			const mockTransactionEffectiveFee = '0.001760';
+			const mockTransactionEffectiveFee = '0.001000';
```
Jaguar0625, line 18: why this change? were tests failing?

cryptoBeliever:
@AnthonyLaw looks good. Big performance improvement for aggregates with multi-inner transactions.

__tests__/infrastructure/TransactionService.spec.js
AnthonyLaw: Because the `payloadSize` affected the fees.

https://github.com/symbol/explorer/blob/21cdac24417aea0bda23dce412ca1fefc2a22d54/__tests__/TestHelper.js#L28

__tests__/infrastructure/CreateTransaction.spec.js
AnthonyLaw: make sense, i will update the description 👍🏼

__tests__/infrastructure/TransactionService.spec.js
Jaguar0625: so, it was failing before?

AnthonyLaw: not, previously the payloadSize was set to `176`.

let me update some of the code.

PR #1116, opened 2022-09-23
[explorer] fix: Restriction display issue

AnthonyLaw:
## What was the issue?
- some formatting issues cause the UI to display data incorrectly.

image

## What's the fix?
- fix and refactor for `RestrictionField` and `RestrictionService`
- added unit test

`/mosaics/1788BA84888894EB`
image

`/accounts/TA4RYHMNHCFRCT2PCWOCJMWVAQ3ZCJDOTF2SGBI`
image

### DIFF
--- a/__tests__/infrastructure/RestrictionService.spec.js
+++ b/__tests__/infrastructure/RestrictionService.spec.js
@@ -1,0 +1,161 @@
+import { RestrictionService } from '../../src/infrastructure';
+import TestHelper from '../TestHelper';
+import { restore, stub } from 'sinon';
+import { MosaicId, UInt64, MosaicRestrictionType, Address } from 'symbol-sdk';
+
+describe('RestrictionService', () => {
+	describe('formatMosaicGlobalRestriction', () => {
+		const runBasicMosaicGlobalRestrictionTests = (restrictionType, expectedResult) => {
+			const createMosaicGlobalRestriction = referenceMosaicId => {
+				return {
+					compositeHash: 'E040D414A6F2E083EC4EBC5060461385E71392F292D6D25ADF7F1CBD0C24CE99',
+					entryType: 1,
+					mosaicId: new MosaicId('1788BA84888894EB'),
+					restrictions: [{
+						key: UInt64.fromUint(790526),
+						referenceMosaicId,
+						restrictionType: restrictionType.type,
+						restrictionValue: UInt64.fromUint(10)
+					}],
+					version: 1
+				};
+			};
+
+			it(`returns mosaic global restriction dto with type ${restrictionType.name} and without reference mosaic`, () => {
+				// Arrange:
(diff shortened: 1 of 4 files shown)


__tests__/infrastructure/RestrictionService.spec.js
```diff
@@ -0,0 +1,161 @@
...
+			it(`returns mosaic global restriction dto with type ${restrictionType.name} and without reference mosaic`, () => {
+				// Arrange:
+				const mosaicGlobalRestriction = createMosaicGlobalRestriction(new MosaicId('0'.repeat(16)));
+
+				// Act:
+				const result = RestrictionService.formatMosaicGlobalRestriction(mosaicGlobalRestriction);
+
+				// Assert:
+				expect(result).toEqual({
+					compositeHash: mosaicGlobalRestriction.compositeHash,
```
Jaguar0625, line 33: title: without reference => with self reference?

or is test wrong in that object contains referenceMosaicId

__tests__/infrastructure/RestrictionService.spec.js
```diff
@@ -0,0 +1,161 @@
...
+				restrictionType: {
+					type: MosaicRestrictionType.GE,
+					name: 'great equal'
+				},
+				expectedResult: 'mosaicRestrictionType.GE'
+			},
+			{
+				restrictionType: {
+					type: MosaicRestrictionType.GT,
+					name: 'greater than'
```
Jaguar0625, line 88: do you intend to mix great and greater ?

cryptoBeliever:
@AnthonyLaw I got a question regarding global restriction. You added mosaicId to the "Restrictions" column:

image

it's possible that we will have another mosaicId that we currently looking at the page?

__tests__/infrastructure/RestrictionService.spec.js
AnthonyLaw: > without reference => with self-reference?

yes you are right. this is better name.

__tests__/infrastructure/RestrictionService.spec.js
AnthonyLaw: yes, I just want to add all MosaicRestrictionType into the test.

AnthonyLaw:
> @AnthonyLaw I got a question regarding global restriction. You added (on beginning) mosaicId to the "Restrictions" column:
> 
> image
> 
> it's possible that we will have another mosaicId that we currently looking at the page?

yes, based on the API, the `restrictions` returns in Array.  

image

Jaguar0625 (approved):

__tests__/infrastructure/RestrictionService.spec.js
Jaguar0625: i mean if you wanted to make the names more distinct, as "greater|less than or equal" would be more proper.

PR #1117, opened 2022-09-24
[explorer] fix: chrome console errors

AnthonyLaw:
## What was the issue?
- mosaic artifact expiry receipt problem, returns incorrectly property cause the errors.

## What's the fix?
- minor refactor on `getMosaicArtifactExpiryReceipt` method, when duration is 0, it should return page info, rather than an empty object.
- fix `getAccountRestrictions` method, it was missing `restrictions` property

Note:  `getAccountRestrictions` unit test has not been added yet, because it required new created file `RestrictionService.spec` in issue #1116


cryptoBeliever (dismissed):

Jaguar0625:
> getAccountRestrictions unit test has not been added yet, because it required new created file RestrictionService.spec in issue https://github.com/symbol/explorer/pull/1116

you should add test after merging 1116

AnthonyLaw:
> > getAccountRestrictions unit test has not been added yet, because it required new created file RestrictionService.spec in issue #1116
> 
> you should add test after merging 1116

Rebased on top of dev and `getAccountRestrictions` unit test

Jaguar0625 (approved):

Issue #1106, opened 2022-09-06
Partial transactions list for account not presented


cryptoBeliever:
Steps to reproduce:
- Init aggregate bonded that require additional signatures from the wallet (for example for 2 of 2 multi-sig) and don't sign it.
- Open account with partial transactions in explorer (transaction will expire but check screenshot):
- In Transaction list component change filter by "Partial" transactions

Result: Transactions are not presented (Transactions component disappear)

https://testnet.symbol.fyi/accounts/TDX2TR7SYUSJTWRGKTYLP6Z64OCZHC3PIOPJYPA

In the console I can see error:
```
Pagination.js:164 TypeError: Cannot read properties of undefined (reading 'length')
    at Function.l [as fromString] (long.js:221:13)
    at Function.isLongNumericString (UInt64.js:74:33)
    at Function.fromNumericString (UInt64.js:62:21)
    at h (CreateTransactionFromDTO.js:95:124)
    at m (CreateTransactionFromDTO.js:199:26)
    at Http.js:125:48
    at Array.map (<anonymous>)
    at l.toPage (Http.js:125:37)
    at TransactionHttp.js:135:83
    at l.<anonymous> (Http.js:107:25)
```

AnthonyLaw:
Due to REST missing the `meta` field from the partial transaction, cause the SDK breaks the explorer.

https://discord.com/channels/856325968096133191/865180563433979904/1016256170996023296

Issue #1105, opened 2022-09-06
Details for partial transaction not presented


cryptoBeliever:
Steps to reproduce:

1. Init aggregate bonded that require additional signatures (for example for 2 of 2 multi-sig) and don't sign it.
2. Open partial transaction in explorer:

Result: Transaction details are not visible

https://testnet.symbol.fyi/transactions/6B9E86105ED6968357436EBD2DFF50D925C8D64F7733D2818688DF7EFAA5F740

![image](https://user-images.githubusercontent.com/17684281/188646513-61f9c8c7-45c5-4f31-bf89-e5af3a5e4853.png)

Expected result:
Explorer should present information about the partial transaction. At this moment we can see only related hash lock.

In the console I can see the error:

```
TypeError: Cannot read properties of undefined (reading 'length')
    at Function.l [as fromString] (long.js:221:13)
    at Function.isLongNumericString (UInt64.js:74:33)
    at Function.fromNumericString (UInt64.js:62:21)
    at h (CreateTransactionFromDTO.js:93:137)
    at CreateTransactionFromDTO.js:186:50
    at Array.map (<anonymous>)
    at m (CreateTransactionFromDTO.js:185:55)
    at TransactionHttp.js:50:115
    at l.<anonymous> (Http.js:107:25)
    at Generator.next (<anonymous>)
```

AnthonyLaw:
Due to REST missing the `meta` field from the partial transaction, cause the SDK breaks the explorer.

https://discord.com/channels/856325968096133191/865180563433979904/1016256170996023296

PR #1127, opened 2022-10-11
[explorer] task: update symbol-sdk to 2.0.2

AnthonyLaw:
## What was the issue?
- Old SDK has a problem handling meta info when the value is undefined.

## What's the fix?
- updated latest SDK 2.0.2

This latest SDK fix issue #1106 #1105

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "merkletreejs": "^0.2.31",
         "moment-timezone": "^0.5.35",
         "nem-sdk": "^1.6.8",
-        "symbol-sdk": "^2.0.1",
+        "symbol-sdk": "^2.0.2",
         "symbol-statistics-service-typescript-fetch-client": "^1.1.5",
         "url-parse": "^1.5.10",
         "vue": "^2.6.14",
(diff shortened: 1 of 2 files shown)


Jaguar0625 (approved):

cryptoBeliever:
@AnthonyLaw issues:
- https://github.com/symbol/explorer/issues/1106
- https://github.com/symbol/explorer/issues/1105
are resolved. 

I just have a question regarding the partial transaction presentation (fee field):
image
I think that for partial transactions we should show a max fee because currently for partial it's always zero (I expect it's calculated as for confirmed tx). If you agree I will raise a separate issue for that.

AnthonyLaw:
> @AnthonyLaw issues:
> 
> * [Partial transactions list for account not presented #1106](https://github.com/symbol/explorer/issues/1106)
> * [Details for partial transaction not presented #1105](https://github.com/symbol/explorer/issues/1105)
> 
> are resolved.
> 
> I just have a question regarding the partial transaction presentation (fee field): image I think that for partial/unconfirmed transactions we should show a max fee because currently for partial it's always zero (I expect it's calculated as for confirmed tx). My proposition:
> 
> * Paid Fee (for confirmed transactions)
> * Max Fee (for unconfirmed/partial transactions)
> 
> If you agree I will raise a separate issue for that.

yes, I think that sounds good to me. let do it

cryptoBeliever:
@AnthonyLaw reported it here -> https://github.com/symbol/explorer/issues/1130

cryptoBeliever (approved):

PR #1129, opened 2022-10-11
[explorer] fix: horizontal scrollbar issue

AnthonyLaw:
## What was the issue?
- Because media screen queries cause the horizontal scrollbar appears when the screen size is big enough. 

## What's the fix?
- We like to keep all columns in the table, so technically the horizontal scrollbar will be helpful on the small screen.
-  I adjusted the `min-width` screen size to 1300px.

### DIFF
--- a/src/components/tables/TableListView.vue
+++ b/src/components/tables/TableListView.vue
@@ -75,26 +75,28 @@
 					</tr>
 				</tbody>
 			</table>
-			<div v-if="pagination || timelinePagination" class="bottom">
-				<div class="pagination-wrapper">
-					<Pagination
-						:canFetchPrevious="prevPageExist"
-						:canFetchNext="nextPageExist"
-						:goUp="false"
-						:currentPageNumber="pageNumber"
-						:lastPageNumber="lastPage"
-						class="pagination"
-						@next="nextPage"
-						@previous="prevPage"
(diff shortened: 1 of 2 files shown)


OlegMakarenko:
Thanks! I can confirm that the horizontal scroll bar has gone from desktop view. But on mobile it still exists. It also moves the pagination controls.

https://user-images.githubusercontent.com/33131259/195305006-e53c6c43-7dd1-4846-981a-bfef339a52e7.mp4

Jaguar0625 (dismissed):

AnthonyLaw:
> Thanks! I can confirm that the horizontal scroll bar has gone from desktop view. But on mobile it still exists. It also moves the pagination controls.
> 
>  Symbol.Block.Explorer.-.Google.Chrome.2022-10-12.12-25-30.mp4

I think the scrollbar is needed, as we want to keep the table and column on small screen

Jaguar0625 (approved):

Jaguar0625 (approved):

OlegMakarenko (approved):
@AnthonyLaw the issue with moving pagination controls is fixed. 👍

https://user-images.githubusercontent.com/33131259/195884732-0d42c9b7-a759-4057-9ab2-c2fd6ef1cffa.mp4

PR #1137, opened 2022-10-17
[explorer] task: hide hash lock section

AnthonyLaw:
## What was the issue?
- hash lock section is displayed on all transaction-type detail pages, technically it should only display in Aggregate Bonded transaction
- hideEmptyData is not working correctly.

## What's the fix?
- Refactoring on transaction store, hash lock info will only request when the transaction type is **AGGREGATE_BONDED**, It will reduce API call and performance increment for other transaction types.
- fix `hideEmptyData` issue.
- added unit test

### DIFF
--- a/__tests__/views/PageAssembler.spec.js
+++ b/__tests__/views/PageAssembler.spec.js
@@ -1,7 +1,5 @@
 import PageAssembler from '../../src/views/PageAssembler.vue';
 import { createLocalVue, shallowMount } from '@vue/test-utils';
-import objects from 'nem-sdk/build/model/objects';
-import { restore } from 'sinon';
 import Vuex from 'vuex';
 
 const localVue = createLocalVue();
@@ -93,11 +91,11 @@ describe('views/PageAssembler', () => {
 
 			runItemShownTests(
 				{
-				    hideOnError: true
+					hideOnError: true
 				},
 				{
-				    error: true
-			    },
+					error: true
+				},
 				false
 			);
 
--- a/__tests__/views/PageAssembler.spec.js
+++ b/__tests__/views/PageAssembler.spec.js
@@ -1,0 +1,115 @@
+import PageAssembler from '../../src/views/PageAssembler.vue';
+import { createLocalVue, shallowMount } from '@vue/test-utils';
+import objects from 'nem-sdk/build/model/objects';
+import { restore } from 'sinon';
+import Vuex from 'vuex';
+
+const localVue = createLocalVue();
+localVue.use(Vuex);
+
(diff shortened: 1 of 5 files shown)


cryptoBeliever:
@AnthonyLaw looks good.

Hash lock information is visible for aggregate bonded (in the partial and confirmed state):
image

Hash lock information is not visible for other transaction types.

cryptoBeliever (approved):

__tests__/views/PageAssembler.spec.js
```diff
@@ -0,0 +1,115 @@
...
+
+			runItemShownTests(
+				{
+				    hideOnError: true
+				},
+				{
+				    error: true
+			    },
+				false
+			);
```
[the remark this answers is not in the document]
AnthonyLaw: hmm, `lint:fix` should fix this issue 🤔

src/views/PageAssembler.vue
```diff
@@ -122,19 +121,23 @@ export default {
...
-			)
-			)
-				return false;
+			if (item.hideEmptyData) {
+				const data = this.getData(item);
+
+				if (Array.isArray(data)) {
+					return !!data.length;
+				} else {
+					return !!Object.keys(data).length
```
OlegMakarenko, line 134: I guess this should work also for arrays, so nested `if` statement is not necessary here.

OlegMakarenko: Also I would suggest to change this part only to return `false` when `hideEmptyData` is true and  `Object.keys(data).length` is 0.
Otherwise when `hideEmptyData` is positive, we will never reach the part below (and all errored views will be rendered):
`[L138] if (item.hideOnError && this.getter(item.managerGetter)?.error)`

__tests__/views/PageAssembler.spec.js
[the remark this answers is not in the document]
AnthonyLaw: ok, I found out that, Explorer didn't have `no-mixed-spaces-and-tabs` in lint.
I will create a separate PR for it.

src/views/PageAssembler.vue
AnthonyLaw: > Also I would suggest to change this part only to return `false` when `hideEmptyData` is true and `Object.keys(data).length` is 0.
> Otherwise when `hideEmptyData` is positive, we will never reach the part below (and all errored views will be rendered):
> `[L138] if (item.hideOnError && this.getter(item.managerGetter)?.error)`

that makes sense 👍🏼

OlegMakarenko (approved):
👍

PR #1152, opened 2022-11-22
[explorer] fix: search namespace without dash

AnthonyLaw:
## What was the issue?
- when the user searches the namespace containing `-`, it not working in explorer, the reason is that the search box is defaulted removed `-`.
- we used to removed `-` to prevent the address with a dash. 

## What's the fix?
- we will keep the `-` in the search box because the namespace allows the string contain a dash.
- we no longer use `-` in our address format (for example wallet), so it won't affect address search.

### DIFF
--- a/__tests__/store/ui.spec.js
+++ b/__tests__/store/ui.spec.js
@@ -131,25 +131,21 @@ describe('store/ui', () => {
 		});
 
 		describe('input value contain whitespace', () => {
-			const runBasicWhitespaceTests = (pageName, inputValue) => {
+			const runBasicWhitespaceTests = (pageName, inputValue, expectedResult) => {
 				it(`returns ${pageName} page`, async () => {
 					// Arrange:
-					if ('mosaic' === pageName) 
+					if ('mosaic' === pageName)
 						stub(MosaicService, 'getMosaicInfo').returns(Promise.resolve({}));
-					 else if('account' === pageName) 
+					else if('account' === pageName)
 						stub(AccountService, 'getAccountInfo').returns(Promise.resolve({}));
-					 else if ('namespace' === pageName) 
+					else if ('namespace' === pageName)
 						stub(NamespaceService, 'getNamespaceInfo').returns(Promise.resolve({}));
-					
 
 					// Act:
 					await ui.actions.search({ dispatch, rootGetters }, inputValue);
 
 					// Assert:
-					expect(dispatch).toHaveBeenNthCalledWith(1, 'openPage', {
-						pageName,
-						param: inputValue.replace(/\s/g, '')
-					});
+					expect(dispatch).toHaveBeenNthCalledWith(1, 'openPage', expectedResult);
 				});
 			};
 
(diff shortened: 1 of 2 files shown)


__tests__/store/ui.spec.js
```diff
@@ -96,23 +110,9 @@ describe('store/ui', () => {
...
+		it('return namespace page with namespace name', async () => await assertSearchNamespace('symbol.xym'));
 
-			// Assert:
-			expect(dispatch).toHaveBeenCalledTimes(1);
-			expect(dispatch).toHaveBeenNthCalledWith(1, 'openPage', {
-				pageName: 'namespace',
-				param: namespaceName
-			});
-		});
+		it('return namespace page with namespace name contain dash', async () => await assertSearchNamespace('symbol-'));
```
Jaguar0625, line 115: should you have test with keyword that includes whitespace ?

AnthonyLaw: I missed that part, let me add it.

AnthonyLaw: I had a minor refactor on the code, hope it make sense

cryptoBeliever (dismissed):

__tests__/store/ui.spec.js
```diff
@@ -131,6 +129,59 @@ describe('store/ui', () => {
...
+						stub(NamespaceService, 'getNamespaceInfo').returns(Promise.resolve({}));
+					
+
+					// Act:
+					await ui.actions.search({ dispatch, rootGetters }, inputValue);
+
+					// Assert:
+					expect(dispatch).toHaveBeenNthCalledWith(1, 'openPage', {
+						pageName,
+						param: inputValue.replace(/\s/g, '')
```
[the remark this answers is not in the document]
AnthonyLaw: Do you mean like rename it?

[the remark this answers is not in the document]
Jaguar0625: no - i mean add an additional parameter instead of `inputValue.replace(/\s/g, '')`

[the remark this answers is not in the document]
AnthonyLaw: patched the fix 👍🏼

__tests__/store/ui.spec.js
```diff
@@ -131,6 +129,58 @@ describe('store/ui', () => {
...
+				},
+				{
+					pageName: 'namespace',
+					input: 'symbol . xym '
+				}
+			];
+
+			inputValues.forEach(({input, pageName}) => runBasicWhitespaceTests(pageName, input, {
+				pageName,
+				param: input.replace(/\s/g, '')
```
Jaguar0625, line 182: consider adding expected to last object and then

param: expected ? expected : input

Jaguar0625 (approved):

__tests__/store/ui.spec.js
AnthonyLaw: > consider adding expected to last object and then

I am confused about this.

the object below is considered as `expectedResult`
```json
{
    pageName,
    param: input.replace(/\s/g, '')
}
```

> param: expected ? expected : input

why do we need this?

PR #1167, opened 2023-02-21
[explorer] fix: unable load peer node in page, and filter only available node in list

AnthonyLaw:
## What was the issue?
- peer/voting node unable load in node page.
- node listing contains stale nodes, it needs to filter out.

## What's the fix?
- filter all the stale nodes, for the node list.
- refactor on getNodeStats, all stats will be calculated based on the available node list.
- refactor on getNodeInfo to fix unable load peer/voting node in page.
- add unit test

### DIFF
--- a/__tests__/TestHelper.js
+++ b/__tests__/TestHelper.js
@@ -366,13 +366,13 @@ const TestHelper = {
 		};
 	},
 	generateNodePeerStatus: isAvailable => {
-		return	{
+		return {
 			isAvailable,
 			lastStatusCheck: 1676809816662
 		};
 	},
 	generateNodeApiStatus: isAvailable => {
-		return	{
+		return {
 			isAvailable,
 			nodePublicKey: '4DA6FB57FA168EEBBCB68DA4DDC8DA7BCF41EC93FB22A33DF510DB0F2670F623',
 			chainHeight: 2027193,
--- a/__tests__/TestHelper.js
+++ b/__tests__/TestHelper.js
@@ -364,6 +364,44 @@ const TestHelper = {
 			cosignatures,
 			innerTransactions
 		};
+	},
+	generateNodePeerStatus: isAvailable => {
+		return	{
+			isAvailable,
+			lastStatusCheck: 1676809816662
+		};
+	},
+	generateNodeApiStatus: isAvailable => {
+		return	{
+			isAvailable,
+			nodePublicKey: '4DA6FB57FA168EEBBCB68DA4DDC8DA7BCF41EC93FB22A33DF510DB0F2670F623',
+			chainHeight: 2027193,
+			finalization: {},
+			nodeStatus: {},
+			restVersion: '2.4.2',
+			restGatewayUrl: 'localhost.com'
+		};
+	},
+	nodeCommonField: {
+		version: 16777989,
+		publicKey: '016DC1622EE42EF9E4D215FA1112E89040DD7AED83007283725CE9BA550272F5',
+		networkGenerationHashSeed: '57F7DA205008026C776CB6AED843393F04CD458E0AA2D9F1D5F31A402072B2D6',
+		port: 7900,
+		networkIdentifier: 104,
+		host: 'node.com',
+		friendlyName: 'node',
+		lastAvailable: '2023-02-19T12:36:04.524Z',
+		hostDetail: {},
+		location: '',
+		ip: '127.0.0.1',
+		organization: '',
+		as: '',
+		continent: '',
+		country: '',
+		region: '',
+		city: '',
+		district: '',
+		zip: ''
 	}
 };
 
--- a/__tests__/TestHelper.js
+++ b/__tests__/TestHelper.js
@@ -376,10 +376,19 @@ const TestHelper = {
 			isAvailable,
 			nodePublicKey: '4DA6FB57FA168EEBBCB68DA4DDC8DA7BCF41EC93FB22A33DF510DB0F2670F623',
 			chainHeight: 2027193,
-			finalization: {},
-			nodeStatus: {},
+			finalization: {
+				height: 2031992,
+				epoch: 1413,
+				point: 7,
+				hash: '6B687D9B689611C90A1094A7430E78914F22A2570C80D3E42D520EB08091A973'
+			},
+			nodeStatus: {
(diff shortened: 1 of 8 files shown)


__tests__/TestHelper.js
```diff
@@ -364,6 +364,53 @@ const TestHelper = {
 			cosignatures,
 			innerTransactions
 		};
+	},
+	generateNodePeerStatus: isAvailable => {
```
Jaguar0625, line 368: there are inline tabs here and in other places, is lint not running properly?

__tests__/components/widgets/NodeStatsWidget.spec.js
```diff
@@ -0,0 +1,121 @@
...
+		it('returns node types stats when data is present', () => {
+			assertNodeStatCounts(
+				{
+					1: 5,
+					2: 5,
+					3: 5,
+					4: 5,
+					5: 0,
+					6: 0,
+					7: 5
```
Jaguar0625, line 106: would be better to have unique numbers to ensure each is indexed correctly

__tests__/infrastructure/NodeService.spec.js
```diff
@@ -0,0 +1,229 @@
...
+			roles: 2,
+			apiStatus: generateNodeApiStatus(false),
+			...nodeCommonField
+		},
+		{
+			roles: 3,
+			peerStatus: generateNodePeerStatus(true),
+			apiStatus: generateNodeApiStatus(false),
+			...nodeCommonField
+		},
```
Jaguar0625, line 29: why does this node not show up in list since it is reachable via peer api?

__tests__/infrastructure/NodeService.spec.js
```diff
@@ -0,0 +1,229 @@
...
+			restVersion: '2.4.2'
+		};
+
+		const expectedChainInfoStatus = {
+			height: 2027193,
+			finalizedHeight: 2031992,
+			finalizationEpoch: 1413,
+			finalizationPoint: 7,
+			finalizedHash:
+				'6B687D9B689611C90A1094A7430E78914F22A2570C80D3E42D520EB08091A973',
```
Jaguar0625, line 181: should fit on one line

__tests__/infrastructure/NodeService.spec.js
```diff
@@ -0,0 +1,229 @@
...
+			const { apiStatus, chainInfo, peerStatus } =
+				await NodeService.getNodeInfo(node.publicKey);
+
+			// Assert:
+			expect(apiStatus).toEqual(expectedResult.apiStatus);
+			expect(chainInfo).toEqual(expectedResult.chainInfo);
+			expect(peerStatus).toEqual(expectedResult.peerStatus);
+		};
+
+		it('returns api node status and chain info when api status is present', async () => {
```
Jaguar0625, line 203: these test names don't seem to match tests, are they reversed ?

__tests__/infrastructure/NodeService.spec.js
```diff
@@ -0,0 +1,229 @@
...
+			await assertNodeStatus(statisticServiceNodeResponse[0], {
+				peerStatus: expectedPeerStatus,
+				apiStatus: {},
+				chainInfo: {}
```
Jaguar0625, line 207: why is there no chainInfo for peer nodes?

__tests__/components/widgets/NodeStatsWidget.spec.js
AnthonyLaw: make sense 👍🏼

__tests__/infrastructure/NodeService.spec.js
AnthonyLaw: The statistic services only provide chain info from the API node.

__tests__/TestHelper.js
AnthonyLaw: > there are inline tabs here and in other places

what do you mean by inline tabs?

>  is lint not running properly?

I didn't get any error when running `npm run lint: fix`

__tests__/infrastructure/NodeService.spec.js
AnthonyLaw: The reason is that the API node is down.

I did the logic check here

https://github.com/symbol/explorer/pull/1167/files#diff-80bd52740de6d63768e32a7ecd0f2980c8f7b135ff2731d2a314cf5b8252c9b4R102-R103

__tests__/infrastructure/NodeService.spec.js
AnthonyLaw: > are they reversed 

yes

> these test names don't seem to match tests

how about this `assertNodeInfoResponse` ?

__tests__/TestHelper.js
Jaguar0625: are you sure lint is configured correctly on test code?
notice it is `return<TAB>{`

__tests__/infrastructure/NodeService.spec.js
Jaguar0625: yes, but if peer is operating it will still be participating in underlying network even if REST is down. so, i'm not sure hiding it is the correct behavior.

__tests__/infrastructure/NodeService.spec.js
Jaguar0625: > how about this assertNodeInfoResponse 

hm? as long as you update the `it` descriptions to match the code, it should be ok

__tests__/infrastructure/NodeService.spec.js
AnthonyLaw: yeah, it makes sense.
I will update the logic.

__tests__/infrastructure/NodeService.spec.js
AnthonyLaw: sorry, I miss reading the message.

__tests__/TestHelper.js
AnthonyLaw: > are you sure lint is configured correctly on the test code?

yes, but not sure why the lint fix does not detect this issue on `return<TAB>{`.

Jaguar0625 (approved):

PR #1194, opened 2024-09-18
[explorer] task: Added convert message to utf8 feature

AnthonyLaw:
## What was the issue?
- The explorer transaction message field didn't support the hex string (drop first 0 bytes in message) for UTF-8.

## What's the fix?
- Added a button, so the user can switch from the default view to UTF8
- It only supports the non-encrypted message and the payload in hex string.

# Screenshoot

Screenshot 2024-09-18 at 3 23 36 PM

image

### DIFF
--- a/__tests__/components/fileds/MessageField.spec.js
+++ b/__tests__/components/fileds/MessageField.spec.js
@@ -106,5 +106,12 @@ describe('MessageField component', () => {
 				payload: '48656C6F2066726F6D2073796D626F6C20736E6170'
 			});
 		});
+
+		it('does not render convert message button when plain message payload is empty', () => {
+			assertConvertFormatButtonNotRender({
+				type: 0,
+				payload: ''
+			});
+		});
 	});
 });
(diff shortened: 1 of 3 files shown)


Wayonb:
Can you fix linting?

Wayonb:
Can you fix the ``apexcharts`` conflict? You will need to delete the package-lock.json to repo. Thanks.

__tests__/components/fileds/MessageField.spec.js
```diff
@@ -0,0 +1,110 @@
...
+		});
+
+		it('does not render convert message format button if encrypted message', () => {
+			assertConvertFormatButtonNotRender({
+				type: 1,
+				payload: '48656C6F2066726F6D2073796D626F6C20736E6170'
+			});
+		});
+	});
+});
```
Wayonb, line 110: Add a newline at the end of the file

src/components/fields/MessageField.vue
```diff
@@ -64,11 +73,33 @@ export default {
 		},
 		title () {
 			return `Click to view ${this.getKeyName(`messageTypeDescriptor_${this.messageType}`)}`;
+		},
+		viewFormatTitle () {
+			return this.isViewAsUTF8 ? `View as default` : `View as utf-8`;
+		},
+		isHexString () {
+			return Convert.isHexString(this.value.payload);
```
Wayonb, line 81: What happens if there is no payload?  I see the convert button when there is no message.

Wayonb (approved):
👍🏾

PR #1195, opened 2024-12-26
[explorer] fix: remove database field in node status with non API roles 

AnthonyLaw:
problem: non-API roles display `Data Base` status in node status.
solution: check on the node roles, `Data Base` only displays node contain API roles.

Before:
image

After:
image

### DIFF
--- a/src/infrastructure/NodeService.js
+++ b/src/infrastructure/NodeService.js
@@ -183,8 +183,6 @@ class NodeService {
 				// Api status
 				formattedNode.apiStatus = {
 					connectionStatus: isAvailable,
-					databaseStatus:
-						'up' === nodeStatus?.db || Constants.Message.UNAVAILABLE,
 					apiNodeStatus:
 						'up' === nodeStatus?.apiNode || Constants.Message.UNAVAILABLE,
 					isHttpsEnabled,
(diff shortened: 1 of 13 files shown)


src/infrastructure/NodeService.js
```diff
@@ -183,8 +183,6 @@ class NodeService {
 				// Api status
 				formattedNode.apiStatus = {
 					connectionStatus: isAvailable,
-					databaseStatus:
-						'up' === nodeStatus?.db || Constants.Message.UNAVAILABLE,
 					apiNodeStatus:
```
Wayonb, line 186: Could you tell me what is an API Node? Light vs full.

src/infrastructure/NodeService.js
```diff
@@ -194,6 +192,14 @@ class NodeService {
 						.format('YYYY-MM-DD HH:mm:ss')
 				};
 
+				// Only API nodes have database status
+				if ([2, 3, 6, 7].includes(node.roles)) {
```
Wayonb, line 196: This works for me now.

Wayonb (changes requested):

src/infrastructure/NodeService.js
AnthonyLaw: `databaseStatus` and `apiNodeStatus` are from the `/node/health` endpoint.

 `/node/health` endpoint response:
```
{"status":{"apiNode":"up","db":"up"}}
```

API node is a full node (roles contain API) because roles contain API, which supports the `/node/health` endpoint.

**Light** rest does not support the `/node/health` endpoint.

Wayonb: Yes, the Light has an API Node with fewer features.  Showing the API Node with an X tells the user it is not working.
I would show it with the green check mark. You could ask  @0x6861746366574  on Discord.

AnthonyLaw: Actually I just worry users confuse, they treat Light REST node as Full REST for development.

I have another idea. 
`Light`: Light Node with the check mark, condition (apiStatus.isAvailable && [1, 3, 5].includes(node.roles) )
`Full`: API node with the check mark (remain the same)

Wayonb (approved):
👍🏾

PR #1198, opened 2025-02-11
[explorer] fix: handle mime empty string in header

AnthonyLaw:
problem: when server hosting a page, mime type repsonse empty string. it cause by after upgrade those dependencies.
solution: Handle mime type on the header reponse in server.


Wayonb (approved):
👍🏾

PR #1203, opened 2025-07-20
[explorer] task: migrate statistics pages to node watch services

AnthonyLaw:
Problem: migrate explorer to node watch services
Solution: 
- Removed everything related library and URL for the statistic services.
- Added query node count endpoint from NodeWatch services.
- Added bubble chart config.
- Convert 2 bar charts (height and finalized height) into one single bubble chart.
- Added more unit tests.

image


Wayonb (approved):
👍🏾

